PR #12, opened 2018-03-14
load da tree

RemoveRusky:
robert is useless


RobertBColton:
You used the built-in item-based tree model from the Qt Framework. LGM uses the built-in item-based tree model from the Java Swing framework. So, you tell me if you think you did the right thing?

If you can't make ResourceModel work as both a tree model and do the data binding, then I suggest creating a custom tree model. You do this by subclassing QAbstractItemModel, which is actually the easier way to manage the tree.
http://doc.qt.io/qt-5/qtwidgets-itemviews-simpletreemodel-example.html

RemoveRusky:
Why do I need to subclass the tree?

RemoveRusky:
In wxwidgets theres an art manager thing. Does qt have something similar?

RobertBColton:
Well, you aren't subclassing the tree, you would be creating your own tree subclassing generic `QAbstractItemModel`. The reason is because it will be easier since we don't have to put our data into their items, we can just return our data.

For example, the `data()` method returns QVariant. You can use the column of the QModelIndex parameter to determine whether to return the QIcon from the map or the QString for the name of the resource/group.

RemoveRusky:
I only need to return a string from the tree though.

RobertBColton:
Right, but when you delete an item in the tree, you also want it to deleted from the tree proto. If you write your own model, then your model can make the delete method do that much easier. Using the build in model, you'll have to listen if an item is deleted and then delete it in the proto, which gets trickier because you have to make sure you are listening to every possible change to the Qt model. Making your own model is like just putting a thin layer of makeup over your data, whereas using the built-in one is like dressing your data up as a full blown tranny and walking it around like one. I don't know if that is clear, I'm just trying to put it into terms Josh would understand.

RemoveRusky:
Nope. I don't see the point. The tree is a string and icon. That's all the data needed. I just need to change the signals for delete and rename in the tree afaik.

ps. y u bever come to irc.

RobertBColton:
I want to fix all of the LGM bugs. I can't find `QIconManager` anymore, I don't know if they deprecated it or what. But it should be being used already by extension of the fact that you are using the resource system:
http://doc.qt.io/qt-5/resources.html

So basically, you just want to put the QIcon's in a map, where the key is the resource kind, like I said and delete the switch statement. Do not delete the icons qrc though because that's also used to alias the icon files and use them in the designer files for the editors. We can later add custom icon packs by allowing the user to specify their own qrc file with the icons inside and load it dynamically. This is better than how LGM does it because in Java the icons are just attached to the jar file in a ZIP (kind of like GM), whereas Qt tools compile them into a form that matches C++'s memory model.

RemoveRusky:
You can never fix all of lgm's bugs. I could use it for 5 min and find 50 more and even if you want to waste your time you can come to irc and answer shit about radialgm at the same time.

RemoveRusky:
also y this project all tabs?

RobertBColton:
It's the Qt default, and yes I prefer tabs. I don't know, they are just easier to work with. When I turn on visible whitespace, in any editor, spaces are just too noisy looking. There's no problem if you configure them correctly (4 spaces wide). I really don't care about the rest of your style, as long as it's good and not ugly like LGM's.

RemoveRusky:
all my editors set to use spaces for enigma's shit. such a pain

RobertBColton:
Well, I don't know, maybe we can switch it to two spaces, but I really don't think I like spaces. Especially two spaces, it just makes everything look really scrunched on my screen.

RemoveRusky:
how I make git ignore submodules when I go to commit? i put /Submodules/ in gitignore but it still show up

RobertBColton:
Remove the final forward slash like in LGM's `.gitignore`:
https://github.com/IsmAvatar/LateralGM/blob/fc06fe2f3caadcaeb713751a2539ad7342a5f065/.gitignore#L6

RemoveRusky:
tried both. still shows changes. How da fux I make a custom tree when you already have default tree in forms? I just want to hook things up not fuck with your ui files.

RobertBColton:
You aren't making a tree, you are calling `ui->treeView->setModel(instanceOfYourCustomModel)`. If you make a custom model then you should use the designer to refactor it into a tree view and not a tree widget. Tree widget uses all the built-in stuffs.

RemoveRusky:
I don't get how I make a data model or why I should bother still.

RobertBColton:
I understand, as I said, it's REALLY hard to explain, you wouldn't be able to see it if you aren't familiar with UI programming or you haven't made one before. It's all about state management, that's it. If you use the built-in tree items and stuff, then you have Qt managing your data. If you write your own model, then you are managing the data, and you are just telling Qt what happens to be in it. So you use a view, for Qt to show what's in your model. The way you have it now is Qt is showing you what is in it's model. By creating a tree item and setting its name, you are duplicating what's in the protos. You are duplicating data, and redundancy is a bad thing.

RobertBColton:
I suggest reading about the concept of a single source of truth:
https://en.wikipedia.org/wiki/Single_source_of_truth

RemoveRusky:
Qt's model does everything I need. I see no need to write my own. If you need a custom model you should write it. I'm fine with the default.

RobertBColton:
I suggest reading about the concept of a single source of truth:
https://en.wikipedia.org/wiki/Single_source_of_truth

> such that every data element is stored exactly once

RemoveRusky:
I'm not intrested in implementing a solution for a problem I don't have.

RobertBColton:
You do have that problem, by setting the name on the tree widget item, you are duplicating data from the protos. You are duplicating data, and that could cause you to later forget about it.

I'll grant you that this is not the least egregious possible way of duplicating the data, but when you consider the rest of this application, redundancy as a strategy is a very bad thing.

RemoveRusky:
There's no need to update the buffer's tree until you save though. I see no need for data model and I won't implement one.

RobertBColton:
Let me clarify further. Your whole strategy of just using signals/slots? It's exactly what LGM does:
https://github.com/IsmAvatar/LateralGM/tree/master/org/lateralgm/ui/swing/propertylink

And that's why LGM has a million bugs with "oh this text field isn't updating", "oh no I'm getting errors typing into this text box"

RemoveRusky:
Let me clarify further. I'm not writing a custom model because I don't need one. If you want one you write it.

RobertBColton:
That's potentially even worse @RemoveRusky, because now you are talking about keeping the UI state and the real data model unsynchronized.

Read this guy, he's genius invented Apple pie or something:
https://martinfowler.com/eaaDev/uiArchs.html

RemoveRusky:
I'm not reading random articles you got from a google search saying "prove my point"

RobertBColton:
Not even if he really is guy who invented Apple pie? Don't you like Apple pie?

RemoveRusky:
No. I like lemon meringue

RobertBColton:
I think @JoshDreamland likes coconut cream...

RemoveRusky:
Anyways, this your last chance to write model or I'm going to continue to do things my way.

RobertBColton:
Ok? That's not a big deal, it doesn't hurt to try both ways. We go with whatever works better honestly. I'm just saying I've already been down your road, and its covered in shit, because its the same shitwagon LGM traveled on.

RemoveRusky:
I'm saying I don't know your grandoise plans and you won't even come to irc to attempt to explain them. So why should I waste my time on them when I can just do it the simple way?

RobertBColton:
I can explain the plan in 3 seconds. You ever draw a UI in a game engine like GM? draw_button(buttonValue)? That's a functional UI style, and it's the basis for immediate mode GUI or IMGUI. Rusky proposed it to me. There's only two ways to do UI, there's that, and then there's Model View Controller, where the button is an object and holds its value separately. Although Qt is MVC, we are trying to design more functional with centralized state where the UI is always current.

RemoveRusky:
nope. I put class BuffGayModel() {]; but no work.

RemoveRusky:
y it not work?

RemoveRusky:
poke

RemoveRusky:
you still havent told me how da fux I write a model. I'ma have to open 500 lgm bugs

RemoveRusky:
error: 'virtual void QTreeWidget::setModel(QAbstractItemModel*)' is private within this context
     ui->treeWidget->setModel(new TreeModel(game->mutable_treeroot(), this));

RemoveRusky:
y it private?

RemoveRusky:
commited a tree model but it doesnt work at all. will you quit hiding and help me with this shit already?

RobertBColton:
What part of it is doesn't work? I looked over it and you seem to be on the right track. I struggled for a couple of hours before I managed to get ResourceModel to show up in the main tree, just as a test, or to do the data binding correctly. I think I was actually missing the flags method, but you seem to have that.

RemoveRusky:
I only see the header and notthing else. When I cout the index method i see some folders listed but theres nothing in the tree displayed

RobertBColton:
I believe line 74 is the problem.

RemoveRusky:
problem how? just come on irc n help me bitch

RobertBColton:
Just try the code I just gave. I'm going to Taco Bell.

RemoveRusky:
works but segfaults when you close window

RobertBColton:
Likely to do with deleting the proto, I don't exactly know how you are supposed to clean up protos other than just deleting them.

RemoveRusky:
y u hiding again? y your model not work? y u not respond on enigma issue?

RemoveRusky:
Hey I pushed a commit trying to start backgrounds. I'm still having issue with the model. when I change the spinbox datachange isnt emited until I click away from it. (You can see this by enabling use as tileset and changing spacing). Any idea how to remedy this?

RemoveRusky:
thanks for the prompt response yet again... anyways, it seems  QEvent::DynamicPropertyChange isn't triggered when you change a checkbox and there's no event that is, other than item clicked.... and even if I update the resource model on that it will change before the checkbox is updated and becomes out of sync....

RobertBColton:
Alright, I worked out all of those issues. What I came up with is actually a much much more technically correct solution, even by Qt standards.

This works bidirectionally, with two editors open, in real time, with immediate commit, and even after closing and reopening. The widgets commit as soon as they are changed, and they don't unnecessarily commit due to style changes. This works under both the Windows style and the Fusion style, unlike my previous solution.

Bidirectional Real Time Databinding

RobertBColton:
Alright, I worked out all of those issues. What I came up with is actually a much much more technically correct solution, even by Qt standards.

This works bidirectionally, with two editors open, in real time, with immediate commit, and even after closing and reopening. The widgets commit as soon as they are changed, and they don't unnecessarily commit due to style changes. This works under both the Windows style and the Fusion style, unlike my previous solution. It also works with the spinboxes, including up/down arrow keys, which I don't think any previous solution did.

Bidirectional Real Time Databinding

RemoveRusky:
why did you change resource model back to a static? it's causing edits one background to effect them all

RobertBColton:
I was just testing, I told you need to make a map anyway. That or make a global ResourceModel.

PR #15, opened 2018-03-23
finish are you sure, finish backgrounds, fix crash on exit

RemoveRusky:


RemoveRusky:
@RobertBColton can you review this when you get up?

RemoveRusky:
this doesn't work if you save an editor then re open and hit don't save changes. I'm not sure why

Models/ProtoModel.cpp
```diff
@@ -1,105 +1,129 @@
...
+
+void ProtoModel::ReplaceBuffer(Message *buffer) {
+  SetDirty(true);
+  protobuf->CopyFrom(*buffer);
+  emit dataChanged(index(0), index(rowCount()));
+}
+
+void ProtoModel::RestoreBuffer() {
+  std::swap(protobuf, protobufBackup);
+  protobufBackup->CopyFrom(*protobuf);
```
RemoveRusky: @JoshDreamland  says should be protobuf->CopyFrom(*protobufBackup) with no swap

JoshDreamland: The important part of what I said is that both will work, but if this way works and the way you just stated does not, something is probably holding a reference longer than it should be, and that could lead to inexplicable crashes down the road. I would investigate this thoroughly.

RobertBColton: Yeah that was the very first thing I tried changing about his code, commenting the swap statement and doing ```protobuf->CopyFrom(*protobufBackup)``` but it did not work. I just tried it again, and it still does not work. Only this version we see here on GitHub will show up unchanged on reopen.

RobertBColton:
I've decided to merge now rather than later because this fixes too many issues on master that I don't want it getting stale. I will have to look deeper into the behavior of the "No" option on the unsaved changes dialog to ensure we aren't leaking memory. I will investigate with Valgrind or another debugging tool.

Models/ProtoModel.cpp
RobertBColton: @JoshDreamland fear not because I came up with a correct fix for it: https://github.com/enigma-dev/RadialGM/commit/b9cae6ef919e27d1850d255c6cf3bf7d6052c53b

The problem was actually just our ImmediateMapper/QDataWidgetMapper forcing the old data back into the widgets because we did not clear the mappings before restoring the buffer (since they are no longer needed) and were not firing a model changed event.

PR #21, opened 2018-04-05
Window State, Classic GMK & YYP Support, and Recent Files

RobertBColton:
This adds the ability to load Classic GameMaker Files and GameMaker: Studio 2's YYP project format from the open dialog.

This follows from the work on https://github.com/enigma-dev/enigma-dev/pull/1179 to get GMK & YYP support in the CLI.

While I was doing this I used QSettings to remember the window state and add a recent files menu.

GMK FPS Example
GM6 Game of Life
GM81 God of Man
GMS2 Arena Shooter

### DIFF
--- a/Components/RecentFiles.cpp
+++ b/Components/RecentFiles.cpp
@@ -1,0 +1,97 @@
+#include "RecentFiles.h"
+
+#include <QFileInfo>
+#include <QSettings>
+#include <QString>
+
+RecentFiles::RecentFiles(QPointer<MainWindow> mainWindow, QPointer<QMenu> menuRecent,
+                         QPointer<QAction> actionClearRecentMenu)
+    : QObject(mainWindow), mainWindow(mainWindow), menuRecent(menuRecent) {
+  menuRecent->connect(menuRecent, &QMenu::aboutToShow, this, &RecentFiles::updateActions);
+  for (int i = 0; i < MaxRecentFiles; ++i) {
+    recentFileActs[i] = menuRecent->addAction(QString(), this, &RecentFiles::recentFileActionTriggered);
+    recentFileActs[i]->setVisible(false);
+  }
+
+  menuRecent->addSeparator();
+  menuRecent->addAction(actionClearRecentMenu);
+
+  menuRecent->setEnabled(this->empty());
+}
+
+static inline QString recentFilesKey() { return QStringLiteral("recentFileList"); }
+static inline QString fileKey() { return QStringLiteral("file"); }
+
+bool RecentFiles::empty() {
+  QSettings settings;
+
+  const int count = settings.beginReadArray(recentFilesKey());
+  settings.endArray();
+  return count > 0;
+}
+
+static QStringList readRecentFiles(QSettings &settings) {
+  QStringList result;
+  const int count = settings.beginReadArray(recentFilesKey());
+  for (int i = 0; i < count; ++i) {
+    settings.setArrayIndex(i);
+    result.append(settings.value(fileKey()).toString());
+  }
+  settings.endArray();
+  return result;
+}
+
+void RecentFiles::updateActions() {
+  QSettings settings;
+
+  const QStringList recentFiles = readRecentFiles(settings);
+  const int count = qMin(int(MaxRecentFiles), recentFiles.size());
+  int i = 0;
+  for (; i < count; ++i) {
+    const QString fileName = QFileInfo(recentFiles.at(i)).fileName();
+    QString numberString = QString::number(i + 1);
+    recentFileActs[i]->setText(numberString.insert(numberString.length() - 1, '&') + " " + fileName);
+    recentFileActs[i]->setData(recentFiles.at(i));
+    recentFileActs[i]->setVisible(true);
+  }
+  for (; i < MaxRecentFiles; ++i) recentFileActs[i]->setVisible(false);
+}
+
+static void writeRecentFiles(const QStringList &files, QSettings &settings) {
+  const int count = files.size();
+  settings.beginWriteArray(recentFilesKey());
+  for (int i = 0; i < count; ++i) {
+    settings.setArrayIndex(i);
+    settings.setValue(fileKey(), files.at(i));
+  }
+  settings.endArray();
+}
+
+void RecentFiles::prependFile(const QString &fileName) {
+  QSettings settings;
+
+  const QStringList oldRecentFiles = readRecentFiles(settings);
+  QStringList recentFiles = oldRecentFiles;
+  recentFiles.removeAll(fileName);
+  recentFiles.prepend(fileName);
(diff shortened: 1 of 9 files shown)


Plugins/ServerPlugin.cpp
```diff
@@ -1,8 +1,9 @@
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600  // at least windows vista required for grpc
```
JoshDreamland: As we discussed, please don't let this get out of hand. Move it somewhere common if it becomes an issue.

MainWindow.cpp
```diff
@@ -49,7 +63,88 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui(new Ui::MainWi
...
+
+bool MainWindow::hasRecentFiles() {
+  QSettings settings;
+
+  const int count = settings.beginReadArray(recentFilesKey());
+  settings.endArray();
+  return count > 0;
+}
+
+static QStringList readRecentFiles(QSettings &settings) {
```
JoshDreamland: Fantastic example of something MainWindow.cpp doesn't need to be in charge of.

MainWindow.cpp
```diff
@@ -49,7 +63,88 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui(new Ui::MainWi
...
+  QSettings settings;
+
+  const QStringList oldRecentFiles = readRecentFiles(settings);
+  QStringList recentFiles = oldRecentFiles;
+  recentFiles.removeAll(fileName);
+  recentFiles.prepend(fileName);
+  if (oldRecentFiles != recentFiles) writeRecentFiles(recentFiles, settings);
+
+  this->ui->menuRecent->setEnabled(!recentFiles.isEmpty());
+}
```
JoshDreamland: You could move all these "recent files" methods to a `RecentFiles.cpp` and miss nothing for it.

MainWindow.cpp
```diff
@@ -108,8 +203,25 @@ void MainWindow::openSubWindow(buffers::TreeNode *item) {
 
 void MainWindow::openFile(QString fName) {
   QFileInfo fileInfo(fName);
-  this->setWindowTitle(fileInfo.fileName() + " - ENIGMA");
-  project = gmx::LoadGMX(fName.toStdString());
+  const QString suffix = fileInfo.suffix();
+
+  if (suffix == "gm81" || suffix == "gmk" || suffix == "gm6" || suffix == "gmd") {
```
JoshDreamland: Another example of "A lot of logic that has nothing to do with the main window." Isn't there already a project manager / file loader that this would fit nicely into?

Plugins/ServerPlugin.cpp
RobertBColton: I won't yet, but I don't think there will ever be a need to. The server can really only have an implementation defined in one location because C++ does not have partial classes. Although it's possible the logic could be separated out into additional server plugin sources if it gets too big, none of those will actually need to include that server proto header. So this is likely the only file that will ever be including GRPC and need that defined.

MainWindow.cpp
RobertBColton: Moved with the other logic for recent files to `Components/RecentFiles.cpp`

MainWindow.cpp
RobertBColton: Moved with the other logic for recent files to `Components/RecentFiles.cpp`

MainWindow.cpp
RobertBColton: No, we don't have sort of `ProjectManager` yet but that if-else ladder could actually go into libEGM since emake has an equivalent if-else ladder. I think I just avoided it because I know fundies was bugging me about using boost for file and directory manipulation since pre-C++14/17 or w/e lacks things like getting the file extension. I did not want to add Boost to libEGM because then the IDE depends on it too, which doesn't make sense because Qt has its own facilities for not just directory and file manipulation, but also command line parsing.

JoshDreamland: Well, all right; I would make this the first thing to go when this file starts creeping up on 500 lines, again.

MainWindow.h
```diff
@@ -57,8 +65,11 @@ class MainWindow : public QMainWindow {
   Ui::MainWindow *ui;
 
   buffers::Project *project;
+  QPointer<RecentFiles> recentFiles;
 
   void openSubWindow(buffers::TreeNode *item);
+  void readSettings();
```
JoshDreamland: Do you intend this to read preferences as well? Right now, it's just reading window geometry/previous run state. So this is kind of a misnomer.

MainWindow.cpp
```diff
@@ -49,7 +53,28 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui(new Ui::MainWi
 
 MainWindow::~MainWindow() { delete ui; }
 
-void MainWindow::closeEvent(QCloseEvent *event) { event->accept(); }
+void MainWindow::readSettings() {
+  QSettings settings;
+
+  settings.beginGroup("MainWindow");
```
JoshDreamland: I only know what the hell this is doing because we just talked about it. Can you add a comment saying, eg,`// Read editor state / window geometry`? Similar for below. Nothing elaborate. The name of this function suggests you're reading user preferences, not just trivia about the way the IDE was last time it was closed.

Components/RecentFiles.cpp
```diff
@@ -0,0 +1,97 @@
+#include "RecentFiles.h"
+
+#include <QFileInfo>
+#include <QSettings>
+#include <QString>
+
+RecentFiles::RecentFiles(QPointer<MainWindow> mainWindow, QPointer<QMenu> menuRecent,
```
JoshDreamland: FYI, this isn't how I pictured this relationship (although I'm fine with it). What I thought you would do is put this file in charge of the recent files list (and negotiating it with the OS), but let Main own the menu and register the listeners. Instead, you're having Main create the menu and then this class registers with it. I would argue that this class should own the menu if it's making fundamental assumptions about the menu existing and how the menu will work. Just my 2¢.

MainWindow.cpp
RobertBColton: Absolutely, like I said, we'll likely figure out libEGM by that time anyway.

MainWindow.cpp
RobertBColton: I've added a comment clarifying it for read and write.

Components/RecentFiles.cpp
RobertBColton: Right, I actually started rewriting it to do that but then was afraid if I put the connect code back into MainWindow you'd disagree so I wasn't sure. I'm still kind of not sure. However, it would simplify it a bit as I wouldn't have to pass as many constructor arguments.

PR #22, opened 2018-04-22
fundies is the real jesus

RemoveRusky:


RobertBColton:
I don't think you want `CONFIG` for this because that is for `*.prf` files which we are not using here. 
http://doc.qt.io/qt-5/qmake-advanced-usage.html
I think we just want a plain old environment variable.

RobertBColton:
Closing because I have included this in #33. This change is technically correct because `DEFINES` is for macros in headers/sources (which we want to ban from this project). And `CONFIG` is for project-wide configuration as it relates to make. So this is good.

Issue #18, opened 2018-03-28
3/28/2018 Tasks


RobertBColton:
- [ ] Read more about the arena allocation and see if we can just swap when restoring the buffer in the proto model: https://developers.google.com/protocol-buffers/docs/reference/arenas
- [ ] Investigate the QHash destruction for subWindows and resourceModels on the main window
- [ ] Add a "heartbeats" option to the CLI server mode that is turned on by default "GRPC_ARG_KEEPALIVE_TIMEOUT_MS"
- [x] Make fundies sprite forms
- [x] Make qscintilla conditional, when off it builds with QPlainTextEdit (so i can debug easier): https://github.com/enigma-dev/RadialGM/commit/b459ba83a5d5305692c829dbe0c487ce26362806
- [x] Investigate ways to change the default layout spacing.
- [x] Review #17

RobertBColton:
Here's some details about me investigating the default layout spacing. I found some different sources.

I found in the QLayout documentation that spacing is actually inherited if not set:
http://doc.qt.io/qt-5/qlayout.html#spacing-prop

RobertBColton:
Side note, the QScintilla MinGW package was fixed for the debug build: https://github.com/Alexpux/MINGW-packages/pull/3507

PR #23, opened 2018-05-02
Sound Editor Proposal

RobertBColton:
This is just a draft of what I think the sound editor should maybe look like. I feel it's a little more organized and compact than LGM's. Most notably, the sound effect options (chorus/flanger/reverb) are missing because ENIGMA hasn't implemented them yet, and they are technically deprecated in GMS because they complicate the sound embedding unnecessarily for a feature that few people need.

The only issue with this pull request perhaps is the default layout margin/spacing issue discussed in #18

Sound Editor Design
Sound Editor Fusion


RobertBColton:
Alright, let's get this in. I'm making a couple of last minute changes here.

* I now resize the new QMdiSubWindow to the size of the editor widget to fix the issue with them all opening too small
* I've disabled the plugin call to test printing of all the JDI discovered resources since it's making the startup really long
* I've removed an unused include to `QMdiSubWindow` in `BaseEditor`
* I'm going to merge without the format or compression options present on the sound editor since we won't be adding the functionality for a while

PR #25, opened 2018-05-07
Preferences Dialog Functionality

RobertBColton:
Alright, so this makes the preferences dialog basically functional and it's modeled after the Qt Creator one. It behaves a little better than LGM's setup because restoring defaults does not do silly things like clear the recent files menu or the window and dock widget state.

* **Restore Defaults**: Clears the "Preferences" group in the QSettings, attempts to set default things like style, then resets the controls
* **Ok**: Accepts the dialog, and when the dialog is accepted it writes the values from the widgets to QSettings then attempts to apply those settings
* **Cancel**: Rejects the dialog, and when the dialog is rejected nothing happens (it's not kept in memory)
* **Apply**: Writes the values from the widgets to QSettings then attempts to apply those settings

There's some problems however.
* Although this is something Qt Creator itself does, technically Mac apps usually apply settings immediately without having to hit an "Apply" button (though I believe GameMaker itself, just like Qt Creator, also ignores this distinction).
* I can't find a good way to reset the style after it's set in main.cpp and I've filed a bug report for it
https://bugreports.qt.io/browse/QTBUG-70325
* I don't know if the organization of all this state is at all logical, Preferences dialogs are usually a pain in the ass like this.

Regardless, this is otherwise working, I am able to save my selected style and reopen RGM with it:
General Preferences
Appearance Preferences


Dialogs/PreferencesDialog.ui
```diff
@@ -147,24 +153,44 @@
            <property name="bottomMargin">
             <number>0</number>
            </property>
-           <item>
-            <widget class="QTabWidget" name="tabWidget">
-             <property name="autoFillBackground">
-              <bool>false</bool>
+           <item row="0" column="1">
```
JoshDreamland: I assume this was generated by something. It's consistent, but it's not very pretty.

MainWindow.cpp
```diff
@@ -201,22 +201,33 @@ void MainWindow::on_actionNext_triggered() { ui->mdiArea->activateNextSubWindow(
 void MainWindow::on_actionPrevious_triggered() { ui->mdiArea->activatePreviousSubWindow(); }
 
 void MainWindow::on_actionDocumentation_triggered() {
-  QUrl documentationURL("https://enigma-dev.org/docs/Wiki/Main_Page", QUrl::TolerantMode);
+  QSettings settings;
+  const QString &documentationURI =
```
JoshDreamland: Confused as to why this object needs to be constructed for every function; might just inline that construction.

Dialogs/PreferencesDialog.cpp
```diff
@@ -1,22 +1,88 @@
...
+    if (style()->objectName().toLower() == styleName.toLower()) ui->styleCombo->setCurrentText(styleName);
+  }
+}
+
+void PreferencesDialog::applyClicked() { this->apply(); }
+
+void PreferencesDialog::restoreDefaultsClicked() {
+  QSettings settings;
+  settings.beginGroup(preferencesKey());
+  settings.remove("");
```
JoshDreamland: I'm not familiar with this idiom; what does removing an empty string do? Apparently, it clears the whole group. I'd comment stuff like this. But then, I guess it might not be unreasonable to expect some prior knowledge... so I won't insist.

Dialogs/PreferencesDialog.ui
RobertBColton: The designer yeah, that thing is difficult to fight too, I've tried fighting the designer on other things too. It loves making a mess out of what you change in the UI file.

MainWindow.cpp
RobertBColton: That's the idiom for QSettings, it's designed that way. The default constructor constructs it with the application name and developer you specified to QApplication.
http://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop

What do you mean by inline it?

Dialogs/PreferencesDialog.cpp
RobertBColton: I'll comment it so that the intention is clear. But yeah, since we are "restoring defaults" we just want to remove the whole group.

Dialogs/PreferencesDialog.cpp
```diff
@@ -1,22 +1,88 @@
...
+static inline QString preferencesKey() { return QStringLiteral("Preferences"); }
+static inline QString generalKey() { return QStringLiteral("General"); }
+static inline QString appearanceKey() { return QStringLiteral("Appearance"); }
+
+void PreferencesDialog::apply() {
+  QSettings settings;
+  settings.beginGroup(preferencesKey());
+
+  settings.beginGroup(generalKey());
+  settings.setValue("documentationURI", ui->documentationLineEdit->text());
```
RobertBColton: @JoshDreamland do you think I should make constants for all of these keys to avoid mistakes/typos causing them to not work? What's a good clean way to do that? I was already doing it for the groups (e.g, appearanceKey()). Is that acceptable?

RobertBColton: I decided to do it anyway: f23ade6188e5314ba8961a15c68e61cea1772f3d

Dialogs/PreferencesDialog.cpp
RobertBColton: I've shortened and used the key helper so that it becomes clear and obvious: https://github.com/enigma-dev/RadialGM/pull/25/commits/4a490abae0b9e7cd73ed45a145c472ca718d42fd#diff-72e7877bb22fe5bbf3b7eb07dfad31ebR75

MainWindow.cpp
RobertBColton: I inlined it in at least one way: 7ec936cdc4199a724fdc8e91e02074defd4b8074

Not sure if it's what you meant, but it eliminates duplication which makes it less error-prone.

Dialogs/PreferencesDialog.cpp
```diff
@@ -1,22 +1,78 @@
...
+          &PreferencesDialog::restoreDefaultsClicked);
+  connect(ui->buttonBox->button(QDialogButtonBox::Apply), &QAbstractButton::clicked, this,
+          &PreferencesDialog::applyClicked);
+}
+
+PreferencesDialog::~PreferencesDialog() { delete ui; }
+
+void PreferencesDialog::apply() {
+  QSettings settings;
+  settings.beginGroup(preferencesKey());
```
JoshDreamland: I don't understand the point of this `begin`/`endGroup` dynamic Qt seems to like, but I really appreciate how you've commented the ends to let readers know what preference they're editing. As discussed over Discord, I think it'd be good if there were some way to recursively reflect into the UI and do this procedural information gathering automatically. But that's an exercise for another time.

Dialogs/PreferencesKeys.h
```diff
@@ -0,0 +1,46 @@
...
+
+inline QString documentationURL() {
+  QSettings settings;
+  QString path = preferencesKey() + "/" + generalKey() + "/" + documentationURLKey();
+  return settings.value(path, "https://enigma-dev.org/docs/Wiki/Main_Page").toString();
+}
+
+inline QString websiteURL() {
+  QSettings settings;
+  QString path = preferencesKey() + "/" + generalKey() + "/" + websiteURLKey();
```
JoshDreamland: I particularly like this path dynamic. Note: it's now "stylish" to overload `operator/` to do this. I don't know why and would not ask you to do that to this codebase...

PR #26, opened 2018-08-26
cmek

RemoveRusky:

PR #28, opened 2018-08-27
Setup AppVeyor Build Using MSVC 2017 & Releases

RobertBColton:
This follows #27 using the instructions from #26 to build RGM on AppVeyor. We just download the Qt static build from `RadialGM-deps` and compile RadialGM and deploy it to the GitHub releases page.


RobertBColton:
How many pdb or lk files are there? Just one? Is it `RadialGM.pdb`?

PR #29, opened 2018-08-31
linux cmek

RemoveRusky:


CMakeLists.txt
```diff
@@ -43,9 +43,12 @@ set(RGM_SOURCES
     Plugins/RGMPlugin.cpp
     Plugins/ServerPlugin.cpp
     Components/RecentFiles.cpp
-    qtplugins.cpp
 )
 
+if (MSVC)
+  set(RGM_SOURCES ${RGM_SOURCES} qtplugins.cpp)
+endif()
```
RobertBColton: I have the same question here as I did on the `enigma-dev` pr. Why don't we make our own static flag so that behavior is not dependent on the compiler?

CMakeLists.txt
```diff
@@ -149,41 +158,44 @@ find_library(LIB_GPR NAMES gpr)
 target_link_libraries(${EXE} PRIVATE ${LIB_GRPC_UNSECURE} ${LIB_GRPC} ${LIB_GPR})
 
 # Find Protobuf
-find_library(LIB_PROTOBUF NAMES libprotobufd)
-find_library(LIB_PROTOC NAMES libprotocd)
-find_library(LIB_PROTOLITE NAMES libprotobuf-lited)
-target_link_libraries(${EXE} PRIVATE ${LIB_PROTOC} ${LIB_PROTOBUF} ${LIB_PROTOLITE})
+include(FindProtobuf)
+include_directories(${Protobuf_INCLUDE_DIRS})
+target_link_libraries(${EXE} PRIVATE ${Protobuf_LIBRARIES})
```
RobertBColton: Same comment here too, why do this for protobuf but not grpc.

RobertBColton: I already answered this over on the enigma-dev PR, there is a `FindGRPC` that was added recently but it's not well documented.

PR #31, opened 2018-09-04
Room Editor Prototype

RobertBColton:
This is just a basic mockup of the room editor to get us started. Basic idea here is to take the best of old GM and modern GM plus any lessons learned from LGM.

Room Editor Settings Tab Fusion Style
Room Editor Views Tab Windows Vista Style

### DIFF
--- a/Editors/RoomEditor.ui
+++ b/Editors/RoomEditor.ui
@@ -259,48 +259,6 @@
            <property name="verticalSpacing">
             <number>4</number>
            </property>
-           <item row="2" column="0">
-            <widget class="QLabel" name="label_15">
-             <property name="text">
-              <string>Width</string>
-             </property>
-             <property name="alignment">
-              <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-             </property>
-            </widget>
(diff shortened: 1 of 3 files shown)


Editors/RoomEditor.ui
```diff
@@ -17,7 +17,1381 @@
...
+             <property name="text">
+              <string>Enemy 2</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>New Item</string>
+             </property>
+            </item>
+            <item>
```
JoshDreamland: This generator fucking sucks. Just saying. This is atrocious, and there's no reason for it.

RobertBColton: Agreed, QML is a lot cleaner, but that requires adding too much more Qt shit onto this. Don't get me wrong, I think QML is really nice, but it requires a whole 'nother engine on top of the other Qt stuff just to render it. Our hands are pretty much tied on this issue, and it doesn't bother me anyway because most of these generators are garbage. Also, I'm just glad this garbage isn't stuffed into the classes like it was in LGM, which makes the actual logic really hard to read.

PR #36, opened 2018-09-27
Recursive Models and Room Editor Rendering

RemoveRusky:
Alright so the goal here was to make it possible to map to submodels easily and this was used to implement some basic room editor functionality.

### Summary of Changes
* Added a 500mb pixmap cache to the `ArtManager` class for caching large images used by the background, sprite, and room renderering widgets. That way they don't each load their own copy. Later it can be expanded with a file change monitor to automatically reload the pixmaps when they are changed on disk.
* Changed ArtManager::GetIcon to load more icons that have not yet been loaded.
* Used `QMenuView` widget by Ulrich Van Den Hekke to implement resource picker.
* Changed `BackgroundEditor` and `ScriptEditor` to use the renamed default mapper of `BaseEditor` which distinguishes it from the mapper used for mapping the resource's name in the tree model.
* Changed `BackgroundEditor::dataChanged` to forward itself to the `BaseEditor` base event.
* Made the `BackgroundEditor` save button mutate the dirty state of the default resource mapper and not the dirty state of its model.
* Changed the `BaseEditor` close event to check the dirty state of the default resource mapper and not the dirty state of its model. When not saving changes it also now clears the mappings to its tree node mapper. Finally, it will otherwise mark its mapper as not dirty now instead of its model.
* Added a map of resource type cases to to tree node field numbers to `BaseEditor.h` to facilitate looking up the submodel of a tree node.
* Added a `ResourceRenamed` signal to `BaseEditor` that is emitted when the tree node of the resource is renamed.
* Used the tree node mapper of `BaseEditor` in the room editor to map the name field on the settings tab.
* Used the resource mapper of `BaseEditor` in the room editor to map the rest of the settings that are fields of the root room resource proto.
* Added a new mapper in the room editor to map the view settings on the views tab.
* Made the current view combo box change the currently mapped view resource on the views tab.
* Tied the selection of assets in the assets list to the model shown in the asset properties view. Later this same selection model needs to be shared with the RoomRenderer as well.
* Added a status bar just under the room renderer that tracks the mouse position.
* Used an event filter to paint the transparency checkerboard pattern on the background of the room preview scroll area viewport.
* Added zooming to the room editor and room renderer.
* Put the room preview and the room property tabs into a horizontal splitter so more screen space can be used when designing levels.
* Because we decided to deprecate the transparent field from GM6 in libEGM when I fixed GMK, I removed the transparent checkbox from the sprite and background editors as well as the logic from both renderers.
* Changed the close event of the main window to close all the subwindows which triggers their save changes dialog. More work needs to be done to this to make it perfect. Even if you decline the individual dialog for each editor, there should still be a final overruling dialog that's shown if the tree model itself is dirty. That is how GM5 behaved though it could be improved upon with "Yes to All" or "No to All".
* Changed the `BaseEditor` constructor to accept the tree node instead of its subresource so editors can map the name field.
* Created the composite `ModelMapper` class that combines `ImmediateDataWidgetMapper` together with `ProtoModel` using traditional class composition. Fundies argued that since everywhere we use a model, we also want a mapper, it ultimately will save redundant code.
* Changed the `ProtoModel` to deal with recursive submodels from nested protobuf messages.
* Added the `RepeatedProtoModel` class for use with repeated submodels, such as instances that are a repeated field in the Room proto.
* Added the `TreeSortFilterProxyModel` class for filtering the main tree by resource type. It is used in conjunction with QMenuView to implement the resource picker which needs to filter by type.
* Added the `RoomRenderer` widget that uses the submodels of the Room proto to render the room layers, views, assets, grid, etc.
* Added the `ResourceModelMap` class that facilitates in finding resources by name as well as getting an object/sprite preview for the tree.
* Fixed a typo in the background renderer's grid painting algorithm. The mistake was that fundies put width where he meant height for the number of horizontal lines.

### Example Rooms
Functional Room Editor
SONIGMA Level0
Game Of Life room0
Key to Success Levels
DragonRise3

### DIFF
--- a/CmakeLists.txt
+++ b/CmakeLists.txt
@@ -41,12 +41,17 @@ set(RGM_SOURCES
     Models/SpriteModel.cpp
     Models/ProtoModel.cpp
     Models/ImmediateMapper.cpp
+    Models/ResourceModelMap.cpp
+    Models/RepeatedProtoModel.cpp
+    Models/TreeSortFilterProxyModel.cpp
     Components/ArtManager.cpp
     Components/Utility.cpp
     Components/RecentFiles.cpp
+    Components/QMenuView.cpp
     Widgets/BackgroundRenderer.cpp
     Widgets/CodeWidget.cpp
     Widgets/ColorPicker.cpp
+    Widgets/RoomRenderer.cpp
     Plugins/RGMPlugin.cpp
     Plugins/ServerPlugin.cpp
     qtplugins.cpp
@@ -72,12 +77,17 @@ set(RGM_HEADERS
     Models/ProtoModel.h
     Models/ImmediateMapper.h
     Models/SpriteModel.h
+    Models/ResourceModelMap.h
+    Models/RepeatedProtoModel.h
+    Models/TreeSortFilterProxyModel.h
     Components/ArtManager.h
     Components/RecentFiles.h
     Components/Utility.h
+    Components/QMenuView.h
     Widgets/BackgroundRenderer.h
     Widgets/CodeWidget.h
     Widgets/ColorPicker.h
+    Widgets/RoomRenderer.h
     Plugins/RGMPlugin.h
     Plugins/ServerPlugin.h
 )
--- a/CmakeLists.txt
+++ b/CmakeLists.txt
@@ -40,6 +40,7 @@ set(RGM_SOURCES
     Models/TreeModel.cpp
     Models/SpriteModel.cpp
     Models/ProtoModel.cpp
+    Models/ModelMapper.cpp
     Models/ImmediateMapper.cpp
     Models/ResourceModelMap.cpp
     Models/RepeatedProtoModel.cpp
@@ -75,6 +76,7 @@ set(RGM_HEADERS
     Editors/SpriteEditor.h
     Models/TreeModel.h
     Models/ProtoModel.h
+    Models/ModelMapper.h
     Models/ImmediateMapper.h
     Models/SpriteModel.h
     Models/ResourceModelMap.h
--- a/Components/QMenuView.cpp
+++ b/Components/QMenuView.cpp
@@ -1,0 +1,247 @@
+/*
+ XINX
+ Copyright (C) 2007-2011 by Ulrich Van Den Hekke
+ xinx@shadoware.org
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful.
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/*
+ * This class is inspired from the Nokia class MenuModel, used in the browser
+ * application in demos.
+ */
+
+// Xinx header
+#include "QMenuView_p.h"
+
+Q_DECLARE_METATYPE(QModelIndex);
+
+/* QMenuViewPrivate */
+
+QMenuViewPrivate::QMenuViewPrivate(QMenuView* menu) : _menu(menu)
+{
+}
+
+QMenuViewPrivate::~QMenuViewPrivate()
+{
+}
+
+
+QAction * QMenuViewPrivate::makeAction(const QModelIndex &index)
+{
+  QIcon icon = qvariant_cast<QIcon>(index.data(Qt::DecorationRole));
+  QAction * action = new QAction(icon, index.data().toString(), this);
+  action->setEnabled(index.flags().testFlag(Qt::ItemIsEnabled));
+  // improvements for Qlipper (petr vanek <petr@yarpen.cz>
+  action->setFont(qvariant_cast<QFont>(index.data(Qt::FontRole)));
+  action->setToolTip(index.data(Qt::ToolTipRole).toString());
+  // end of qlipper improvements
+  QVariant v;
+  v.setValue(index);
+  action->setData(v);
+
+  return action;
+}
+
+
+void QMenuViewPrivate::triggered(QAction *action)
+{
+  QVariant v = action->data();
+  if (v.canConvert<QModelIndex>())
+  {
+    QModelIndex idx = qvariant_cast<QModelIndex>(v);
+    emit _menu->triggered(idx);
+  }
+}
+
+void QMenuViewPrivate::hovered(QAction *action)
+{
+  QVariant v = action->data();
+  if (v.canConvert<QModelIndex>())
+  {
+    QModelIndex idx = qvariant_cast<QModelIndex>(v);
+    QString hoveredString = idx.data(Qt::StatusTipRole).toString();
+    if (!hoveredString.isEmpty())
+      emit _menu->hovered(hoveredString);
+  }
+}
+
+void QMenuViewPrivate::aboutToShow()
+{
+  QMenu * menu = qobject_cast<QMenu*>(sender());
+  if (menu)
+  {
+    QVariant v = menu->menuAction()->data();
+    if (v.canConvert<QModelIndex>())
+    {
+      QModelIndex idx = qvariant_cast<QModelIndex>(v);
+      _menu->createMenu(idx, menu, menu);
+      disconnect(menu, SIGNAL(aboutToShow()), this, SLOT(aboutToShow()));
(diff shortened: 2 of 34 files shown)


RobertBColton:
I am just making this comment so we don't forget to solve some things before merging this. At the current time I have discovered and rooted out the following issues.

1) After updating the enigma-dev submodule to pull my event name fixes, you now need to copy `events.res` next to the RadialGM exe or it will crash loading certain projects.
2) Projects that have no instances in a room will crash because we are attempting to read the model. This is sort of ok for now, because we need to add layers next and then obviously fix the crash if there's nothing in the layer.
3) The room renderer needs to cache the images it is painting with or painting tiles can be so slow the window never opens.

Editors/RoomEditor.cpp
```diff
@@ -1,13 +1,141 @@
...
 
-  QGraphicsScene* scene = new QGraphicsScene(this);
-  ui->graphicsView->setScene(scene);
+  ProtoModel* roomModel = model->GetSubModel(TreeNode::kRoomFieldNumber);
+  ui->roomRenderer->SetResourceModel(roomModel);
+
+  nodeMapper->addMapping(ui->roomName, TreeNode::kNameFieldNumber);
+  nodeMapper->toFirst();
+
+  resMapper->addMapping(ui->speedSpinBox, Room::kSpeedFieldNumber);
```
JoshDreamland: This is the cruft I was suggesting getting rid of using attributes or properties or whatever on the Qt XML. I think you mentioned some snag you hit doing that in Discord, but I don't remember it. If it was just that these sources are precompiled, that doesn't actually matter. If you just put a `protofield` attribute in the UI elements, you could automatically look them up in a generalized loop.

Editors/RoomEditor.cpp
```diff
@@ -1,13 +1,141 @@
...
+  ui->objectSelectButton->setMenu(objMenu);
+
+  connect(objMenu, &QMenu::triggered, this, &RoomEditor::SelectedObjectChanged);
+
+  connect(ui->currentViewComboBox, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged),
+          [=](int index) { viewMapper->setCurrentIndex(index); });
+
+  RepeatedProtoModel* m = roomModel->GetRepeatedSubModel(Room::kInstancesFieldNumber);
+  ui->layersAssetsView->setModel(m);
+  connect(ui->layersAssetsView->selectionModel(), &QItemSelectionModel::selectionChanged,
```
JoshDreamland: This code just maps a combobox to a proto enum, right? This seems sufficiently general to be moved elsewhere.

Editors/RoomEditor.cpp
```diff
@@ -1,13 +1,141 @@
...
+          [=](const QItemSelection& selected, const QItemSelection& /*deselected*/) {
+            if (selected.empty()) return;
+            auto selectedIndex = selected.indexes().first();
+            auto currentInstanceModel =
+                roomModel->data(Room::kInstancesFieldNumber, selectedIndex.row()).value<void*>();
+            ui->layersPropertiesView->setModel(static_cast<ProtoModel*>(currentInstanceModel));
+          });
+
+  for (int c = 0; c < m->columnCount(); ++c) {
```
JoshDreamland: This also looks general... but I can't tell what it's for.

MainWindow.cpp
```diff
@@ -105,34 +108,23 @@ void MainWindow::openSubWindow(buffers::TreeNode *item) {
...
 
-    if (!resourceModels.contains(item)) resourceModels[item] = new ProtoModel(typeMessage, nullptr);
-    auto resourceModel = resourceModels[item];
+    ProtoModel *res = resourceMap->GetResourceByName(item->type_case(), item->name());
+    BaseEditor *editor = factoryFunction->second(res, this);
 
-    QWidget *editor = factoryFunction->second(resourceModel, nullptr);
-    resourceModel->setParent(editor);
+    connect(editor, &BaseEditor::ResourceRenamed, resourceMap, &ResourceModelMap::ResourceRenamed);
+    connect(editor, &BaseEditor::ResourceRenamed, [=]() { treeModel->dataChanged(QModelIndex(), QModelIndex()); });
```
JoshDreamland: I *like* this.

CmakeLists.txt
```diff
@@ -40,13 +40,19 @@ set(RGM_SOURCES
     Models/TreeModel.cpp
     Models/SpriteModel.cpp
     Models/ProtoModel.cpp
+    Models/ModelMapper.cpp
```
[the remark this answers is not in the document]
RobertBColton: I wouldn't know that reason and this is actually getting painful for me to update both lists of sources in the qmake pro and the cmake lists, but I would like to continue to support building with both. Mainly because I use qtcreator when writing pull requests even though we are building releases with MSVC.

Components/QMenuView.cpp
```diff
@@ -0,0 +1,247 @@
+/*
+ XINX
```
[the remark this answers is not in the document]
RobertBColton: Yes it is third party and the file's history has only a single commit: https://github.com/enigma-dev/RadialGM/commits/fd9c943d8d70b4283fca9f8f70471e0d25946563/Components/QMenuView.cpp

So, yes, I believe the code is unmodified.

Editors/RoomEditor.cpp
RobertBColton: Yes, I said we could maybe do that using dynamic properties. But this is more complicated now that we have begun mapping nested fields like with rooms.
http://doc.qt.io/archives/qt-4.8/designer-widget-mode.html#dynamic-properties

Let me say first it's very obvious we could do something like string "views/visible" path properties to the protofield of each widget. But I hate this for two reasons, we would have to recurse over the widget hierarchy to look up this dynamic property to make the mapping. Also, it's not quite as fail safe as our mapper code which is checked at compile time (though warnings on the dynamic mapping would make it easier to detect an incorrect mapping).

The problem with QDataWidgetMapper and our own derived ImmediateMapper is that they are mainly oriented at master-detail interfaces. Hence why we are having to create new mappers for each submodel/subproto message.

Editors/RoomEditor.cpp
RobertBColton: Sort of, it maps your selection in the instance list to which instance is displayed in the asset properties view. You know, you select an instance and it shows its properties. Later, I actually want this selection model shared with the room renderer too, like LGM. This setup is a bit like master detail... or just like Qt Designer for instances instead of widgets. It could be moved into a helper yes, but it's not quite needed anywhere else just yet.

That said, it's interesting that you raised this question. I just realized I would like to deal with multiple selections differently. You see, in Qt Creator, if you select multiple widgets then change a property in its property inspector thingy, then it changes it for all selected widgets that have that common property. I think our instances property inspector should work the same way. That makes it easier to set the property for a bunch of instances at once. If you want to set a single instances property, then just select only that instance.

Editors/RoomEditor.cpp
RobertBColton: This is just fundies trying to hide certain columns. Because the properties view shows the more detailed instance properties, he doesn't want the instance list showing them all either. He just wants the instance list to show name/id/sprite. I admit, it could be generalized, and he may have actually already generalized most of it.

MainWindow.cpp
RobertBColton: Which part of it? If you mean the factory, that was me, he tacked on this resource renaming stuff. That said, I think there's a mistake here. Where he passes `QModelIndex()` to fire `dataChanged` I believe that's saying that the entire tree model changed when really only a single node did.

Editors/RoomEditor.cpp
JoshDreamland: It's not going to get easier, and it's not going to be a one-size fits all solution. If you can do half the IDE like that, you'll save *thousands* of lines of useless boilerplate code. Had I done this project on my own, I'd probably have used wx and written my own YAML-based UI parser with these sorts of features. Qt gives us something workable out of the box. Use it where you can.

MainWindow.cpp
JoshDreamland: I like the simple way in which you have registered a handler for a resource rename.

CmakeLists.txt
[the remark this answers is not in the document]
JoshDreamland: You can just `glob()` the directories you want. I think there's even an exclude mechanism so you can colocate test sources with the same extension.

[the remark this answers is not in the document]
RobertBColton: Fundies says don't glob, because cmake ppl annoying about it and will cry. C++ sucks at deps, we can come back to this later.

Components/QMenuView.cpp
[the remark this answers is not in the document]
RobertBColton: fundies says push it on the qt team, you say send a pr to the original bigass it came from, I cba.

PR #37, opened 2018-10-12
GRPC & Compiler Related Features

RobertBColton:
There are two main goals of this pull request. I was hoping to mock up a nice way of selecting configurations and targets and make compiling work. Additionally I wanted to get completions working for the script editor.

Compiling Games

Summary of Changes
----
* Added `SettingsEditor` designer form class with the basic ENIGMA settings for compiling.
* Added 4 new icons to use in the script editor's autocompletion window. One of them is an f(x) icon for functions and the other three are colored right-facing arrows for constants/variables/globals. Resources will later just use their resource icons when we add that.
* Added a system cache to the main window for plugins to communicate available systems to the built in editors. This is primarily just used by the settings editor so it can populate the API combos using systems found by the GRPC plugin.
* Added a static instance accessor to the main window I'd like to later remove that is used to emit a config change signal from outside the main window itself. The reason is that signals require a QObject instance.
* Added a `CompileStatusChanged` signal to RGMPlugin interface so plugins can inform the main window when the compile status changes. This is how running the game disables the other buttons and forces the output dock widget visible.
* Added a static mutator to set the currently selected config which for now just fires a signal on the static instance that I want to remove. This is used temporarily for the settings editor to have a save button that actually tells the GRPC plugin to tell emake to update its API targets.
* Added the settings editor to the editor factory map so config resources can actually be opened.
* Replaced the change game settings action with a menu for config selection under the resources menu in the same place.
* Added a config selector to the main toolbar and gave it a split popup mode.
* Renamed the timeline actions in the menu to be a single word instead of two words. I never agreed with GameMaker making them two words, unlike every other resource group in the tree.
* Added mnemonics to the resource creation menu to facilitate keyboard navigation.
* Merged `ErrorRead` and `OutputRead` signals into a `LogOutput` signal on the main window. This is to support more granularity of logging which we can later allow the user to filter.
* Moved the initialization of the `TreeModel` icon map to its constructor to allow public access. This is so that later the icons can be used to represent resource keywords in the autocompletions.
* Added `future_deadline` helper to facilitate in calculating GRPC deadlines in the `ServerPlugin`.
* Added `tag`/`detag` helpers to the `ServerPlugin` to faciliate in tagging RPC states with async GRPC.
* Added `CallData` to manage the common state of async RPC calls in the `ServerPlugin`.
* Added `AsyncReadWorker` to drive the state machine of an async streaming read RPC in the `ServerPlugin`.
* Added `AsyncResponseReadWorker` to drive the state machine of a non-streaming async read RPC in the `ServerPlugin`.
* Changed all RPC stub implementations in `CompilerClient` to async using derivations of the above classes, helpers, and abstractions.
* Added `CompilerClient::ScheduleTask` to schedule an RPC to be started and processed asynchronously.
* Added `CompilerClient::UpdateLoop` for processing RPC calls asynchronously one at a time (for now). This is triggered regularly by a QTimer with an interval of 0 ms.
* Added a list of search paths for emake that looks for executable files in known locations. This means RGM can now find emake if it's directly next to it in the filesystem.
* Changed the `ServerPlugin` destructor to call `QProcess::close` instead of `QProcess::terminate` because the latter does not work for console programs that do not run an event loop on Windows.
* Added a static keyword API to `CodeWidget` that can be used to communicate keywords to QScintilla for syntax highlighting and autocompletion.

Change Details
----
There is only a single drop down for the current config and target options. Clicking on the button will open the game settings for the currently selected config.
Edit Selected Config

Selecting the arrow will bring up the menu to select a different config, edit the selected one, or add a new one. There is no option to delete a config because I did not want to overcomplicate this menu. Because configs will be a regular resource group, they can be deleted from the main project tree.
Change Selected Config

Finally, the exact same menu will be displayed under the "Resources" menu. The main reason is not just because it's consistent with how GM always was, but just to facilitate keyboard navigation using mnemonics.
Change Settings from Resources Menu

The ENIGMA settings will be integrated with the settings editor.
ENIGMA API Settings

Extensions appear here as well, as we have come to expect.
ENIGMA Extension Settings

### DIFF
--- a/Editors/SettingsEditor.cpp
+++ b/Editors/SettingsEditor.cpp
@@ -1,9 +1,29 @@
 #include "SettingsEditor.h"
+#include "MainWindow.h"
 #include "ui_SettingsEditor.h"
+
+#include <QDebug>
 
 SettingsEditor::SettingsEditor(ProtoModel* model, QWidget* parent)
     : BaseEditor(model, parent), ui(new Ui::SettingsEditor) {
   ui->setupUi(this);
+
+  const QMap<QString, QComboBox*> systemUIMap = {
+      {QString("Audio"), ui->audioCombo},          {QString("Platform"), ui->platformCombo},
+      {QString("Graphics"), ui->graphicsCombo},    {QString("Widget"), ui->widgetsCombo},
+      {QString("Collision"), ui->collisionsCombo}, {QString("Compilers"), ui->compilerCombo},
+      {QString("Network"), ui->networkingCombo},
+  };
+  foreach (auto system, MainWindow::systemCache) {
+    const QString systemName = QString::fromStdString(system.name());
+    auto it = systemUIMap.find(systemName);
+    if (it == systemUIMap.end()) continue;
+    foreach (auto subsystem, system.subsystems()) {
+      const QString subsystemName = QString::fromStdString(subsystem.name());
+      auto combo = it.value();
+      combo->addItem(subsystemName);
+    }
+  }
 }
 
 SettingsEditor::~SettingsEditor() { delete ui; }
--- a/Editors/SettingsEditor.cpp
+++ b/Editors/SettingsEditor.cpp
@@ -4,6 +4,7 @@
 
 #include "codegen/Settings.pb.h"
 
+#include <QDebug>
 #include <QPushButton>
 
 using namespace buffers::resources;
@@ -14,16 +15,33 @@ SettingsEditor::SettingsEditor(ProtoModel* model, QWidget* parent)
 
   QPushButton* saveButton = ui->buttonBox->button(QDialogButtonBox::Save);
   saveButton->setIcon(QIcon(":/actions/accept.png"));
+  connect(saveButton, &QPushButton::clicked, [=]() {
+    qDebug() << "hey";
+    Settings settings;
+    auto* api = settings.mutable_api();
+    api->set_target_audio(ui->audioCombo->currentData().toString().toStdString());
+    api->set_target_collision(ui->collisionCombo->currentData().toString().toStdString());
+    api->set_target_compiler(ui->compilerCombo->currentData().toString().toStdString());
+    api->set_target_graphics(ui->graphicsCombo->currentData().toString().toStdString());
+    api->set_target_network(ui->networkCombo->currentData().toString().toStdString());
+    api->set_target_platform(ui->platformCombo->currentData().toString().toStdString());
+    api->set_target_widgets(ui->widgetsCombo->currentData().toString().toStdString());
+    api->add_extensions("Paths");
+
+    qDebug() << "hi" << ui->audioCombo->currentData().toString();
+
+    emit MainWindow::setCurrentConfig(settings);
+  });
   QPushButton* discardButton = ui->buttonBox->button(QDialogButtonBox::Discard);
   discardButton->setIcon(QIcon(":/actions/cancel.png"));
 
   treePageMap = {{QString("ENIGMA/API"), ui->apiPage}, {QString("ENIGMA/Extensions"), ui->extensionsPage}};
 
   const QMap<QString, QWidget*> systemUIMap = {
-      {QString("Audio"), ui->audioCombo},          {QString("Platform"), ui->platformCombo},
-      {QString("Graphics"), ui->graphicsCombo},    {QString("Widget"), ui->widgetsCombo},
-      {QString("Collision"), ui->collisionsCombo}, {QString("Compilers"), ui->compilerCombo},
-      {QString("Network"), ui->networkingCombo},   {QString("Extensions"), ui->extensionsList},
+      {QString("Audio"), ui->audioCombo},         {QString("Platform"), ui->platformCombo},
+      {QString("Graphics"), ui->graphicsCombo},   {QString("Widget"), ui->widgetsCombo},
+      {QString("Collision"), ui->collisionCombo}, {QString("Compilers"), ui->compilerCombo},
+      {QString("Network"), ui->networkCombo},     {QString("Extensions"), ui->extensionsList},
   };
   foreach (auto system, MainWindow::systemCache) {
     const QString systemName = QString::fromStdString(system.name());
@@ -40,14 +58,16 @@ SettingsEditor::SettingsEditor(ProtoModel* model, QWidget* parent)
     }
     foreach (auto subsystem, system.subsystems()) {
       const QString subsystemName = QString::fromStdString(subsystem.name());
+      const QString subsystemId = QString::fromStdString(subsystem.id());
       const QString subsystemDesc = QString::fromStdString(subsystem.description());
-      if (combo)
-        combo->addItem(subsystemName);
-      else if (listWidget) {
+      if (combo) {
+        combo->addItem(subsystemName, subsystemId);
+      } else if (listWidget) {
         auto item = new QListWidgetItem(subsystemName, listWidget);
         item->setFlags(item->flags() | Qt::ItemFlag::ItemIsUserCheckable);
         item->setCheckState(Qt::Unchecked);
         item->setToolTip(subsystemDesc);
+        item->setData(Qt::UserRole, subsystemId);
       }
     }
   }
--- a/Editors/SettingsEditor.cpp
+++ b/Editors/SettingsEditor.cpp
@@ -53,7 +53,7 @@ SettingsEditor::SettingsEditor(ProtoModel* model, QWidget* parent)
       {QString("Collision"), ui->collisionCombo}, {QString("Compilers"), ui->compilerCombo},
       {QString("Network"), ui->networkCombo},     {QString("Extensions"), ui->extensionsList},
   };
-  foreach (auto system, MainWindow::systemCache) {
+  for (auto system : MainWindow::systemCache) {
     const QString systemName = QString::fromStdString(system.name());
     auto it = systemUIMap.find(systemName);
     if (it == systemUIMap.end()) continue;
@@ -74,7 +74,7 @@ SettingsEditor::SettingsEditor(ProtoModel* model, QWidget* parent)
         ui->systemDesc->setPlainText(subsystemDesc);
       });
     }
-    foreach (auto subsystem, system.subsystems()) {
+    for (auto subsystem : system.subsystems()) {
       const QString subsystemName = QString::fromStdString(subsystem.name());
       const QString subsystemId = QString::fromStdString(subsystem.id());
       const QString subsystemDesc = QString::fromStdString(subsystem.description());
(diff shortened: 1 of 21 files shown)


Hughtech:
I like the mock-up. 

 My only question is, will be a short description pop-up when you mouse over the Extensions for a couple of secs? if not then I think it would be an well-added feature.

RobertBColton:
Good idea. I've done it just for the extensions for now. The reason being is because they are the only system type where all the descriptions are short enough to fit into a single line tooltip without word wrapping. However, we could consider a "brief" field alongside "description" in each system's yaml descriptor that allows you to provide a brief description short enough for tooltips. It'll be slightly more complicated maybe to add this for the combo selected systems.
Extension Tooltips

Widgets/CodeWidgetScintilla.cpp
```diff
@@ -1,27 +1,60 @@
...
+void CodeWidget::finalizeKeywords() { sciApis->prepare(); }
+
+CodeWidget::CodeWidget(QWidget* parent) : QWidget(parent), font(QFont("Courier", 10)) {
+  prepare_scintilla_apis();
+
+  QFontMetrics fontMetrics(font);
 
   QsciScintilla* codeEdit = new QsciScintilla(this);
   this->textWidget = codeEdit;
+  codeEdit->registerImage(0, QPixmap(":/actions/right-arrow-red.png"));
```
JoshDreamland: What's going on with the hard-coded numbers? What do they mean? Why isn't this just happening in some kind of loop? Does nothing cache these images? Why are object and sprite assumed to be pre-loaded, but those other two images are not? Don't you have some kind of icon manager for this?

Widgets/BackgroundRenderer.cpp
```diff
@@ -70,7 +70,7 @@ void BackgroundRenderer::paintEvent(QPaintEvent * /* event */) {
     painter.fillRect(QRectF(0, 0, pixmap.width() * zoom, pixmap.height() * zoom), ArtManager::GetTransparenyBrush());
 
     painter.scale(zoom, zoom);
-    bool transparent = model->data(Background::kTransparentFieldNumber).toBool();
+    bool transparent = true;  // deprecated
```
JoshDreamland: Why true? Isn't the default false, or does this mean "image is already transparent"?

Plugins/RGMPlugin.h
```diff
@@ -12,13 +15,14 @@ class RGMPlugin : public QObject {
   ~RGMPlugin();
 
  signals:
-  void OutputRead(const QString &output);
-  void ErrorRead(const QString &error);
+  void LogOutput(const QString &output);
```
JoshDreamland: Isn't this supposed to have a severity attached? I'll assume you haven't gotten to that, yet.

Editors/SettingsEditor.cpp
```diff
@@ -0,0 +1,110 @@
...
+    api->set_target_audio(get_combo_system_id(ui->audioCombo));
+    api->set_target_collision(get_combo_system_id(ui->collisionCombo));
+    api->set_target_compiler(get_combo_system_id(ui->compilerCombo));
+    api->set_target_graphics(get_combo_system_id(ui->graphicsCombo));
+    api->set_target_network(get_combo_system_id(ui->networkCombo));
+    api->set_target_platform(get_combo_system_id(ui->platformCombo));
+    api->set_target_widgets(get_combo_system_id(ui->widgetsCombo));
+    api->add_extensions("Paths");
+
+    qDebug() << "hi" << ui->audioCombo->currentData().toString();
```
JoshDreamland: Well this is tacky. 😛

Editors/SettingsEditor.cpp
```diff
@@ -0,0 +1,110 @@
...
+             {"controls", ui->controlsPage}, {"graphics", ui->graphicsPage},     {"project info", ui->projectInfoPage},
+             {"version", ui->versionPage}};
+
+  const QMap<QString, QWidget*> systemUIMap = {
+      {QString("Audio"), ui->audioCombo},         {QString("Platform"), ui->platformCombo},
+      {QString("Graphics"), ui->graphicsCombo},   {QString("Widget"), ui->widgetsCombo},
+      {QString("Collision"), ui->collisionCombo}, {QString("Compilers"), ui->compilerCombo},
+      {QString("Network"), ui->networkCombo},     {QString("Extensions"), ui->extensionsList},
+  };
+  foreach (auto system, MainWindow::systemCache) {
```
JoshDreamland: Is there a reason to use this macro in 2018?

Editors/SettingsEditor.cpp
```diff
@@ -0,0 +1,110 @@
...
+    : BaseEditor(model, parent), ui(new Ui::SettingsEditor) {
+  ui->setupUi(this);
+
+  QPushButton* saveButton = ui->buttonBox->button(QDialogButtonBox::Save);
+  saveButton->setIcon(QIcon(":/actions/accept.png"));
+  connect(saveButton, &QPushButton::clicked, [=]() {
+    qDebug() << "hey";
+    Settings settings;
+    auto* api = settings.mutable_api();
+    api->set_target_audio(get_combo_system_id(ui->audioCombo));
```
JoshDreamland: This is an extremely common pattern in this IDE that could be automated with those custom properties. Guess it's not a big deal, since at least it's clearly divided into modules.

MainWindow.cpp
```diff
@@ -45,17 +52,32 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui(new Ui::MainWi
...
+  settingsButton->setPopupMode(QToolButton::ToolButtonPopupMode::MenuButtonPopup);
+  settingsButton->setToolButtonStyle(Qt::ToolButtonStyle::ToolButtonTextBesideIcon);
+  ui->actionSettings->setMenu(ui->menuChangeGameSettings);
+
   RGMPlugin *pluginServer = new ServerPlugin(*this);
   auto outputTextBrowser = this->ui->outputTextBrowser;
-  connect(pluginServer, &RGMPlugin::OutputRead, outputTextBrowser, &QTextBrowser::append);
-  connect(pluginServer, &RGMPlugin::ErrorRead, outputTextBrowser, &QTextBrowser::append);
+  connect(pluginServer, &RGMPlugin::LogOutput, outputTextBrowser, &QTextBrowser::append);
+  connect(pluginServer, &RGMPlugin::CompileStatusChanged, [=](bool finished) {
```
JoshDreamland: I'd move this lambda into the compile plugin territory, or something, as an actual function. That said, I'm pleased this MainWindow source is only a few hundred lines; I was afraid it would grow without bound.

Plugins/ServerPlugin.cpp
```diff
@@ -1,124 +1,285 @@
...
+  virtual void started() {}
+  virtual void finished() {}
+  virtual void process(const T& data) = 0;
+};
 
-  void CompileBuffer(Game* game, CompileMode mode) {
-    QTemporaryFile t("enigma");
-    if (!t.open()) return;
-    CompileBuffer(game, mode, t.fileName().toStdString());
+struct ResourceReader : public AsyncReadWorker<Resource> {
```
JoshDreamland: Some of this logic is general and works for any kind of gRPC server. Some of it is specific to, eg, the compiler plugin. Why not separate the ENIGMA service code into an ENIGMA service header/source?

Widgets/CodeWidgetScintilla.cpp
RobertBColton: 1) The numbers thing is QSci's way of tagging keywords, it's the QSci analogy to grpc's tags.
2) Whatever you want them to mean.
3) uwutm8
4) I don't know if anything cache's them, my code doesn't, but QsciScintilla might.
5) I just hacked two resources on there to demonstrate it real fast, I don't actually have it hooked up to show resource keywords just yet. We actually also need a custom derived GML/EDL lexer in order to highlight these keywords, which is pretty easy (just override the keyword method and return them etc).
6) We have ArtManager which I added caching of pixmaps to in the other room editor pr.

But yeah, I'll take some steps to clean this shit up, I think we could use an enum too.

Widgets/BackgroundRenderer.cpp
RobertBColton: That's a mistake, thanks for catching it, I did mean false.

Plugins/RGMPlugin.h
RobertBColton: I haven't gotten to it yet 😋

Editors/SettingsEditor.cpp
RobertBColton: Yeah, I'll remove all this temporary debug printing before this is ever merged.

Editors/SettingsEditor.cpp
RobertBColton: Not really no, and its biggest downfall is that foreach makes a copy of the container.

Editors/SettingsEditor.cpp
RobertBColton: Sure, but this one is actually going away now that I have recursive mappers from merging the room editor pr. It will be replaced with the mapper stuff, which still isn't your ideal fix, but will be better.

MainWindow.cpp
RobertBColton: Well the idea behind having the lambda here is because if somebody makes a plugin for a different compiler, say Dejavu, then the behavior should work right out of the box as far as the editing state the IDE is in goes (e.g, the Run button being enabled or disabled). But yeah, I don't expect the main window class to get much bigger either. We've definitely done a better job of organizing here than LateralGM.

Plugins/ServerPlugin.cpp
RobertBColton: What you actually mean is to separate out the generic grpc stuff into its own translation unit. It's not quite as easy as you think because, if you remember, last night I added a `LogOutput` signal to CallData and made it a `QObject` so that it can communicate stuff back through to the main window. That's sort of easy to change though, but, we don't really need to worry about this just yet because there's no real need to until the time someone wants to integrate Dejavu or something. Not really a priority for me anyway.

Widgets/CodeWidgetScintilla.cpp
RobertBColton: Alright: 4c2f5a33d5a03ce4234b8e9a7a1a341b95e3c07c

I've typed the keywords now with an enum which should make that a lot more sensible. There's some work left to be done here. One thing to do is to derive a custom QSciLexer from QSciLexerCPP for GML/EDL so these keywords are also syntax highlighted and things like `with` are considered keywords. We also need to include the resource keywords which is going to be tricky because it doesn't seem possible to update keywords once they are added and prepared. You also apparently can't have more than one `QSciApi` associated with a lexer.

MainWindow.cpp
RobertBColton: Also, let me add that I did it this way so I didn't have to expose the run/debug/compile action/buttons to plugins.

Editors/SettingsEditor.cpp
RobertBColton: Alright: 10dd3ae97b9fe667019119934dcf90ee8158afae

It's all gone now, it's not needed now anyway.

PR #39, opened 2018-10-20
Basic Color Picking Widget

RobertBColton:
Well, yeah, we need one of these so here it is. I obviously used a `Q_PROPERTY` to implement the user property of the widget so that it can be mapped to models with a widget mapper. It handles transparency just fine using the transparency pattern and drawing a filled rectangle with the color on top of that. Went ahead and promoted the one in the room editor to it.

Room Grid Color Picking

### DIFF
--- a/Widgets/ColorPicker.cpp
+++ b/Widgets/ColorPicker.cpp
@@ -12,19 +12,20 @@ ColorPicker::ColorPicker(QWidget *parent) : QWidget(parent) {
   this->setLayout(layout);
 
   connect(button, &QToolButton::clicked, [=]() {
-    QColor color = QColorDialog::getColor(this->color_, this, tr("Choose a color"), QColorDialog::ShowAlphaChannel);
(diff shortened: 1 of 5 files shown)


Widgets/ColorPicker.cpp
```diff
@@ -0,0 +1,37 @@
...
+  connect(button, &QToolButton::clicked, [=]() {
+    QColor color = QColorDialog::getColor(this->color_, this, tr("Choose a color"), QColorDialog::ShowAlphaChannel);
+    if (!color.isValid()) return;
+    this->setColor(color);
+  });
+  connect(this, &ColorPicker::colorChanged, this, &ColorPicker::updateButtonIcon);
+  this->updateButtonIcon();
+}
+
+QColor ColorPicker::color() const { return this->color_; }
```
JoshDreamland: No need for this `this->`. Also in three places below.

RobertBColton: Fair enough: 1575b9f1ae7f877cf2e97a9cda43da857f8eec84

PR #40, opened 2018-10-29
Generalize Asset Views

RobertBColton:
I just moved the common grid painting algorithm into `Components/Utility.h` for now. This isn't too many less lines of code right now, but later the function will be needed elsewhere. The sprite import dialog will need it to show the splicing of subimages for example.


RobertBColton:
Ok, up for review again, I've redone it by creating `AssetView` as a base class of both the Room and Background views and moved the grid painting code there instead of the `Utility.h` header.

Widgets/AssetView.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const qreal& AssetView::GetZoom() const { return zoom; }
+
+void AssetView::paintGrid(QPainter& painter, int width, int height, int gridHorSpacing, int gridVertSpacing,
+                          int gridHorOff, int gridVertOff, int gridWidth, int gridHeight) {
+  painter.setCompositionMode(QPainter::RasterOp_SourceXorDestination);
+  painter.setPen(QColor(0xff, 0xff, 0xff));
+
+  if (gridHorSpacing != 0 || gridVertSpacing != 0) {
```
JoshDreamland: `&&`, or else handle when when or the other is 0

JoshDreamland: I misread your variables; you aren't handling the case where `gridWidth` / `gridHeight` are zero. Is that handled elsewhere?

RobertBColton: Ok: a1efab0188be39b1a8a2ffe9eec39a0b954a1615

So fundies did it the way he did it because of half-pixel issues with QPainter/QPen. It draws the line off by a half a pixel. Regardless, fundies way is an inefficient way of dealing with that so I did change the logic the way you wanted.

JoshDreamland: LGTM—another option to consider is just not drawing grid lines with spacing less than some predetermined constant (eg, 3), which would handle the 0 case naturally.

RobertBColton:
Sure, but the thing is that still doesn't stop some hoodlum from calling this function on a zero-area grid. That said, I am going to merge this when it passes the CI.

RobertBColton:
Alright, I've done a few more tweaks and I have it pretty much perfect now. The half pixel alignment in Qt is pretty confusing. What you have to do is count the number of tile cells to make sure it's the same as GM8 and that's what I've done here.

* Create AssetView base class
* Rename BackgroundRenderer to BackgroundView
* Rename RoomRenderer to RoomView for consistency
* Move the zoom member into the base AssetView class
* Change the tileset properties spinboxes to use the same min (0) and max (32000) as GM5 and GM8
* Return early and do not paint zero-area grids
* Added comments explaining why certain things are done.
* Use a pen with `Qt::MiterJoin` to prevent the edges of the rectangle-based grid from having diamond-shaped holes.
* Use a pen with `Qt::SquareCap` which gives half-pixel (actually half pen width) extension so the spots where the lines of the line-based grid overlap and the xor cancels is not 50/50 transparent or opaque.
* Offset the non-zero grid spacing by half a pixel for correct alignment.
* Save and restore the state of the `QPainter` before returning so it doesn't affect the caller.

RobertBColton:
Alright: 4adc36aaead9951f0d38cf8a0b8499acf06b918b

I've done a few more tweaks and I have it pretty much perfect now. The half pixel alignment in Qt is pretty confusing. What you have to do is count the number of tile cells to make sure it's the same as GM8 and that's what I've done here.

* Create AssetView base class
* Rename BackgroundRenderer to BackgroundView
* Rename RoomRenderer to RoomView for consistency
* Move the zoom member into the base AssetView class
* Change the tileset properties spinboxes to use the same min (0) and max (32000) as GM5 and GM8
* Return early and do not paint zero-area grids
* Added comments explaining why certain things are done.
* Use a pen with `Qt::MiterJoin` to prevent the edges of the rectangle-based grid from having diamond-shaped holes.
* Use a pen with `Qt::SquareCap` which gives half-pixel (actually half pen width) extension so the spots where the lines of the line-based grid overlap and the xor cancels is not 50/50 transparent or opaque.
* Offset the non-zero grid spacing by half a pixel for correct alignment.
* Fixed the offset of the line-based grid by starting its painted lines at the horizontal or vertical offset. I prefer doing it this way as opposed to just translating the entire painter for both grid methods because it can short circuit and not have to paint the entire grid in certain cases.
* Save and restore the state of the `QPainter` before returning so it doesn't affect the caller.

RobertBColton:
Alright: 4adc36aaead9951f0d38cf8a0b8499acf06b918b

I've done a few more tweaks and I have it pretty much perfect now. The half pixel alignment in Qt is pretty confusing. What you have to do is count the number of tile cells to make sure it's the same as GM8 and that's what I've done here.

* Create AssetView base class
* Rename BackgroundRenderer to BackgroundView
* Rename RoomRenderer to RoomView for consistency
* Move the zoom member into the base AssetView class
* Change the tileset properties spinboxes to use the same min (0) and max (32000) as GM5 and GM8.
* Disabled wrapping on the tileset spinboxes so you can't go below the min (0) and get the max (32000) because nobody will ever want the max for these properties.
* Return early and do not paint zero-area grids
* Added comments explaining why certain things are done.
* Use a pen with `Qt::MiterJoin` to prevent the edges of the rectangle-based grid from having diamond-shaped holes.
* Use a pen with `Qt::SquareCap` which gives half-pixel (actually half pen width) extension so the spots where the lines of the line-based grid overlap and the xor cancels is not 50/50 transparent or opaque.
* Offset the non-zero grid spacing by half a pixel for correct alignment.
* Fixed the offset of the line-based grid by starting its painted lines at the horizontal or vertical offset. I prefer doing it this way as opposed to just translating the entire painter for both grid methods because it can short circuit and not have to paint the entire grid in certain cases.
* Save and restore the state of the `QPainter` before returning so it doesn't affect the caller.

PR #42, opened 2018-10-31
Creating Resources/Project Switching

RobertBColton:
This is a small one but I am just trying to cover a bunch of ground that needs covered.

* Moved `MainWindow::resourceMap` member from a raw pointer to a `QScopedPointer` with explicitly no parent to prevent double delete and ensure cleanup when switching projects.
* Moved `MainWindow::treeModel` member from a raw pointer to a `QScopedPointer` with explicitly no parent to prevent double delete and ensure cleanup when switching projects.
* Moved some logic from `MainWindow::openFile` to a new method  `MainWindow::openProject` so the same logic can be used to implement the new project action and to ensure that the same logic is used to reset the tree and model and such. The new method also closes all of the open subwindows before switching the project, which is necessary and what LGM and GM have always done too.
* Added `MainWindow::openNewProject` to make the new project action actually work and so the logic would not need duplicated for the main window constructor to initially start with a new project. The method also creates the new project with default folder/groups for all of our supported resource types.
* NOTE: `openNewProject` is called from the main window constructor so that if a project to load was specified on the command line and it fails, the window will still be initialized with a new project and be in a usable state.
* Added `MainWindow::CreateResource` which creates a new resource of the passed type case using reflection. The method uses the `MainWindow::resourceMap` to find a unique name though that implementation could later be made more efficient by having the resource map store the maximum/last id by type case. The new resource is added to the root tree node for now and then the resource is opened for editing.
* Added `ResourceModelMap::AddResource` to add newly created resources without having to duplicate logic from its constructor.
* Added `TreeModel::addNode` to add new nodes to the tree.


MainWindow.cpp
```diff
@@ -181,27 +183,52 @@ void MainWindow::openFile(QString fName) {
...
   }
 
   MainWindow::setWindowTitle(fileInfo.fileName() + " - ENIGMA");
   recentFiles->prependFile(fName);
+  openProject(loadedProject);
+}
+
+void MainWindow::openNewProject() {
+  MainWindow::setWindowTitle(tr("<new game> - ENIGMA"));
+  auto newProject = new buffers::Project();
```
JoshDreamland: 1. Try to put `*` after `auto` anywhere you know it's a pointer.
2. Why isn't this a `unique_ptr`? In general, please don't say `new` anywhere in the codebase.

RobertBColton: Sorry about (1) that was a mistake, I intended to do that. As far as (2) it is a unique_ptr now on MainWindow, but I haven't switched libegm to that yet because I didn't want to make conflicts (gmk/gmx have a unique_ptr they are just releasing it before return; don't ask me why I did that it was silly of me to create this technical debt, I know), I am planning to do this soon. Regardless, that's why the `MainWindow::openProject` takes a raw pointer. However, I can make a commit that fixes half of it, give me a second.

RobertBColton: Alright: 1a99871cb3b7ec0b44627a7b4a79c9b41093ffbb

I'll get to having all the libegm load methods returning a unique_ptr soon.

PR #45, opened 2018-11-05
Fix Crash When Opening Unknown Resources

RobertBColton:
This was a minor logic mistake that I made. I used the subscript operator `[]` to check if a subwindow was already added for a given tree node. This was not what I actually intended, because the subscript operator will insert the item if it's not present. The other place that the subscript operator is still used was intentional.

1) Attempt to open an unknown resource by double clicking, like the help/info node
2) Select the help/info tree node with a single left click and then press F2 to rename it
3) After renaming the item the crash occurs

This basically fixes a crash that occurs when trying to open unknown resources (those with type case not set). One such resource at the moment in the help node I've been addressing in other ways. Regardless of whether we want to keep help/info as a resource, this change still needs to be made for safety. This basically means opening an unknown resource is a noop.

### DIFF
--- a/MainWindow.cpp
+++ b/MainWindow.cpp
@@ -132,8 +132,8 @@ void MainWindow::openSubWindow(buffers::TreeNode *item) {
                                 {TypeCase::kSettings, EditorFactory<SettingsEditor>}});
 
   auto swIt = subWindows.find(item);
-  QMdiSubWindow *subWindow = (swIt == subWindows.end()) ? nullptr : *swIt;
-  if (subWindow == nullptr) {
+  QMdiSubWindow *subWindow = *swIt;
+  if (swIt == subWindows.end() || !subWindow) {
     auto factoryFunction = factoryMap.find(item->type_case());
     if (factoryFunction == factoryMap.end()) return;  // no registered editor
 
--- a/MainWindow.cpp
+++ b/MainWindow.cpp
@@ -132,8 +132,8 @@ void MainWindow::openSubWindow(buffers::TreeNode *item) {
                                 {TypeCase::kSettings, EditorFactory<SettingsEditor>}});
 
   auto swIt = subWindows.find(item);
-  QMdiSubWindow *subWindow = nullptr;
-  if (swIt == subWindows.end() || subWindow == nullptr) {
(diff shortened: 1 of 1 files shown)


MainWindow.cpp
```diff
@@ -131,8 +131,9 @@ void MainWindow::openSubWindow(buffers::TreeNode *item) {
                                 {TypeCase::kRoom, EditorFactory<RoomEditor>},
                                 {TypeCase::kSettings, EditorFactory<SettingsEditor>}});
 
-  auto subWindow = subWindows[item];
-  if (!subWindows.contains(item) || subWindow == nullptr) {
+  auto swIt = subWindows.find(item);
+  QMdiSubWindow *subWindow = (swIt == subWindows.end()) ? nullptr : *swIt;
+  if (subWindow == nullptr) {
```
JoshDreamland: Just check if `swIt == end()`, here, and then assign `subWindow = *swIt` below. Unless `*swIt` can be null? That would be pretty ugly. But I'd just include that in the if, here.

RobertBColton: You're missing the part outside the closing brace. `*swIt` can be null and this logic will still work but it won't/shouldn't be. We need `subWindow` whether it's found or not, in order to show it and set it active. When it isn't found, we attempt to open it, so in either case we have a valid `subWindow` by that point.

https://github.com/enigma-dev/RadialGM/blob/a30230a4702f34f07a8f6f605ab618b9049bd6c2/MainWindow.cpp#L154

JoshDreamland: What I'm advising is that you check both null (`== end`) and missing (`!*it`) in this conditional, instead of doing a ternary on the line before to quietly handle part of this check. After the if, `*swIt` will be a valid pointer.

RobertBColton: Alright: 74979a0e67e8af343e49921ed512bf6697eabfb1

How's that? I tested it again, seems to be correct.

RobertBColton: Ok: 2f6aef530c127ef32c4975a63876414b056eb35b

You didn't like that, so you said to do this. Alright? We good now?

JoshDreamland: Yes, thank you. 😛 Fire away.

PR #46, opened 2018-11-11
Resource Renaming Fix

RobertBColton:


MainWindow.cpp
```diff
@@ -150,7 +155,7 @@ void MainWindow::openSubWindow(buffers::TreeNode *item) {
     subWindow->connect(subWindow, &QObject::destroyed, [=]() { subWindows.remove(item); });
 
     subWindow->setWindowIcon(subWindow->widget()->windowIcon());
-    subWindow->setWindowTitle(QString::fromStdString(item->name()));
+    editor->setWindowTitle(QString::fromStdString(item->name()));
```
JoshDreamland: I'm a little confused about this. Sort of odd to do all this work on one window, and then just switch to another to set the title. What's the title of subwindow? Are you making assumptions about the presentation code when setting this?

MainWindow.cpp
```diff
@@ -231,18 +236,8 @@ void MainWindow::openProject(std::unique_ptr<buffers::Project> openedProject) {
...
-                           auto index = topLeft.sibling(row, column);
-                           buffers::TreeNode *item = static_cast<buffers::TreeNode *>(index.internalPointer());
-                           if (!subWindows.contains(item)) return;
-                           auto subWindow = subWindows[item];
-                           subWindow->setWindowTitle(QString::fromStdString(item->name()));
-                         }
-                       }
-                     });
+  treeModel->connect(treeModel.get(), &TreeModel::ResourceRenamed, resourceMap.get(),
+                     &ResourceModelMap::ResourceRenamed);
```
JoshDreamland: Thank you. Please try to avoid ten-line lambdas. RGM's use of them has become... somewhat excessive.

MainWindow.cpp
RobertBColton: This change here is what allowed me to remove the lambda you thanked me for just below in your other comment. You can tell that although Qt is cross-platform it is heavily influenced by Win32 because every QWidget has a window title and icon and can be turned into a full blown window.
http://doc.qt.io/qt-5/qwidget.html#windowTitle-prop

So anyway, by setting it on the editor which is just a QWidget that we pass to `addSubWindow` instead of directly on the subwindow, we can have the editor set its own window title. By making `BaseEditor` capable of setting its own window title (which it does in data changed when the resource is renamed) we are able to eliminate the lambda from below.

MainWindow.cpp
RobertBColton: Right, I don't know, we've just sort of used them where it's easier. I've perhaps been abusing them a bit. Having looked this over again, I may later change this completely around again later. One reason to change it again is so the editor subwindow also has the same icon as the tree node.

PR #47, opened 2018-11-12
Add Tree Node Crash Fix

RobertBColton:
This fixes a crash that can be reproduced by creating a room and then an object. The logic mistake I made was in `TreeModel::addNode` where the child size is one larger than the index of the row I want to insert into the tree. The reason is because the child is already added to the parent by way of `TreeNode::add_child` which is called in `MainWindow::CreateResource` before `addNode` is even called.

I will need to be extra careful when I do the tree drag and drop.


RobertBColton:
You know, perhaps we will set up unit tests when I go to do the tree drag and drop next.

PR #54, opened 2018-12-26
Edit/Tree Context Menu

RobertBColton:


Models/TreeModel.cpp
```diff
@@ -172,6 +180,23 @@ QMimeData *TreeModel::mimeData(const QModelIndexList &indexes) const {
...
+  auto insertIndex = parent;
+  if (!parent.isValid()) insertIndex = QModelIndex();
+  auto *parentNode = static_cast<buffers::TreeNode *>(insertIndex.internalPointer());
+  if (!parentNode) {
+    insertIndex = QModelIndex();
+    parentNode = root;
+  }
+  beginInsertRows(insertIndex, row, row);
+  parentNode->mutable_child()->AddAllocated(node);
+  for (int j = parentNode->child_size() - 1; j > row; --j) {
```
JoshDreamland: You could do this with a truncate and append, but that wouldn't really matter... I can't believe this is essentially the best way of inserting elements. Can you extract this logic to a template to make it easier to replace when Proto gets its shit straight?

Models/TreeModel.cpp
```diff
@@ -220,35 +245,80 @@ bool TreeModel::dropMimeData(const QMimeData *mimeData, Qt::DropAction action, i
...
 
-void TreeModel::addNode(buffers::TreeNode *child, buffers::TreeNode *parent) {
-  auto rootIndex = QModelIndex();
-  emit beginInsertRows(rootIndex, parent->child_size(), parent->child_size());
-  parent->mutable_child()->AddAllocated(child);
-  parents[child] = parent;
-  emit endInsertRows();
+QModelIndex TreeModel::addNode(buffers::TreeNode *child, const QModelIndex &parent) {
+  auto insertIndex = parent;
+  buffers::TreeNode *parentNode = nullptr;
```
JoshDreamland: Please move this declaration into the parent.isValid branch. It's confusing to keep it around when it's barely ever applicable (it seems as though it's the parent you'll be inserting into, but it isn't used outside that branch except where an inline is preferable).

If it makes you feel better, rename `insertIndex` to `insertParent` or similar.

Models/TreeModel.h
```diff
@@ -36,7 +36,11 @@ class TreeModel : public QAbstractItemModel {
   bool dropMimeData(const QMimeData *data, Qt::DropAction action, int row, int column,
                     const QModelIndex &parent) override;
 
-  void addNode(buffers::TreeNode *child, buffers::TreeNode *parent);
+  void insert(const QModelIndex &parent, int row, buffers::TreeNode *node);
```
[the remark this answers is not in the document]
RobertBColton: Good point: dadc4b76d5a69442d298fd58c059e10878f83229

I agree that change makes it more sensible.

Models/TreeModel.cpp
RobertBColton: Alright: 1ec23f6c28c5526fe55663c50cf89c5e6fcd2925

That seems reasonable enough.

Models/TreeModel.cpp
RobertBColton: Alright: 37be6d6242030e7b2aa3b35d1bd5fa6c3bae5780

That's a good point, I was actually going to do this before but decided to wait. Also fundies did his own thing in `SpriteModel.cpp` because he's inserting more than just a single item. We'll address that later though because we need to come back to the sprite editor and finish up still.

PR #55, opened 2019-02-19
Update ENIGMA Submodule

RobertBColton:
We've made a number of changes to enigma-dev recently that affect RadialGM's build and I wanted to get them hammered out now that enigma-dev has calmed down. The only real conflict here was just that I had to change the `protos` directory from `CommandLine` to `shared` where Josh moved them. Also had to include some other tweaks, because I guess we need `grpc_cronet` and `Wtsapi32` now.


RobertBColton:
It seems we're getting OpenSSL errors now like Rusky was getting. So, looks like I'm going to have to go knock on some doors.

PR #58, opened 2019-04-18
update submodule & fix error in cmakelists

fundies:


RobertBColton:
The file `qtplugins.cpp` still exists and it's still very much necessary for the static build. So, what are you doing?
https://github.com/enigma-dev/RadialGM/blob/f1ebe6e716e256d458ce11848a30d31d0f7cc134/qtplugins.cpp#L1

fundies:
it's an if for msvc. one of us just forgot to remove it from general src list

RobertBColton:
Thanks.

PR #59, opened 2019-05-05
Update ENIGMA Submodule

RobertBColton:
Lots of problems today with helping daz and Bloodseeker build for Windows and Linux respectively. Some of the problems were caused by the submodule falling behind and losing fixes Josh made to prevent `JavaStruct.h` pollution into the protos. Also with the newer Qt Creator, it decided to give me some `*.stackdump` file we obviously don't want to be committing, so I've added that to ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ ui_*.h
 *.qmlc
 *.jsc
 *build-*
+*.stackdump
 
 # Qt unit tests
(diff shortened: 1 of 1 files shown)


fundies:
looks like you have the same pr from a much more talented dev opened 17 days ago

RobertBColton:
Thanks for the reminder.

Issue #62, opened 2019-05-09
Model Index Validity Checking


RobertBColton:
I want to talk about #61 which was just merged. What daz did there does in deed fix the crash when creating a new sprite by checking if the model index is valid in the sprite model. However, it did not fix where our code was actually invalid, which is in the sprite editor.
https://github.com/enigma-dev/RadialGM/blob/8fdcf5fd5c018c18fc782a91c9b0366a422a3fee/Editors/SpriteEditor.cpp#L15

What I think we need to consider here is if we should be checking if the model indexes are valid in our data models at runtime. This is what is done in official Qt examples. However, I am concerned it will allow us to get away with writing buggy/incorrect code and have it going unnoticed. What I think should be considered is that we move these model index validity checks to the debug build and change them to assertions.

dazappa:
OK if we are confident we should always pass valid indexes to the models, I like your idea. If you fix the underlying issue of why that index was invalid, I could probably go through all the models and add the asserts to the model::data functions.

RobertBColton:
Ok, Rusky basically did reach the same conclusion I did. However, now I think I've changed my opinion. I think I want to actually do this an unconventional way.

Being that this is a C++ IDE, and naturally more prone to crashes than LateralGM ever would be. I think I am in favor of [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming). And I think we should turn the valid checks into assertions for debug mode. If there's an easy way for us to get that behavior, without turning all of these checks into four lines, I think that's what we should do.

I am going to wait and run this yet by Josh.

JoshDreamland:
The IDE is already heavily invested in Google technology; consider using a logger. Alternatively, consider using your own logging system. When you're supposed to get a valid ID, but do not, log it. 

Surface these error messages in a debug console. Have an icon that opens the debug console. When it has errors in it, change the icon to include a red dot, possibly including a number. When the issue is severe, display a message (like in LGM). Offer Help→Report a problem, and have it include a summary of the log in the suggested bug report.

Caveat: This log could fill up FAST. Mitigate problems with enormous quantities of error messages. Reduce duplicate log messages by digesting them to `{ first timestamp, last timestamp, count }`. Use the source file + line number to identify these. *Include that in the log.*

RobertBColton:
[...]

PR #63, opened 2019-05-10
Defensive Editor Diagnostics

RobertBColton:
This pull request is the product of the discussion in #62. There is a mutual consensus that we should use defensive programming conventions in RGM. This includes logging warnings in release mode and debug mode. Here I have added such defensive logging for model indexes provided to most of our model classes.

The `qWarning` macro was used for logging with no use of assertions. I don't see it as necessary to force the program to exit when it is capable of recovering, and adding all of these as assertions too will just make this diff even fatter. We have made the decision to reserve these Qt macros for internal RGM issues only. If we later decide to reuse QDebug for controlling the build output verbosity, then we plan to use a special instance of the QDebug class.

Me and Josh collaborated on `Components/Logger.h` to make this schizophrenic precondition checking less redundant. It's actually becoming more standard practice in modern years, see Mozilla Firefox.

### Toggle Editor Diagnostics
Toggle Editor Diagnostics
### Example Editor Diagnostics
Example Editor Diagnostics


Models/ProtoModel.cpp
```diff
@@ -128,6 +135,10 @@ QVariant ProtoModel::data(int row, int column) const {
 }
 
 QVariant ProtoModel::data(const QModelIndex &index, int role) const {
+  if (!index.isValid()) {
+    qWarning() << index << "provided to ProtoModel::data for role" << role << "is invalid.";
```
JoshDreamland: I love how every strange, far-fetched thing I tell you to do, Qt has a builtin for.

RobertBColton: Actually that's kind of funny now that you mention it. Kind of a bit of a bitter sweet irony I'd say.

JoshDreamland: No, it just means that Qt is keeping up with the trends. This sort of logger is becoming prevalent, and it's nice to see it included in the framework. It makes a ton of sense for the UI framework to just handle this, and it saves us a bunch of code. All while still giving us access to the information I wanted (eg, source file + line number). It's great.

RobertBColton: Sure, I get that, I just thought you were being tongue in cheek too. But ok, great!

Models/ProtoModel.cpp
```diff
@@ -78,6 +81,10 @@ bool ProtoModel::IsDirty() { return dirty; }
 int ProtoModel::columnCount(const QModelIndex & /*parent*/) const { return 1; }
 
 bool ProtoModel::setData(const QModelIndex &index, const QVariant &value, int role) {
+  if (!index.isValid()) {
+    qWarning() << index << "provided to ProtoModel::setData for role" << role << "with value" << value << "is invalid.";
```
JoshDreamland: You don't have spaces around those values. Or any others. How can Qt know to separate them? And how does it delimit them?

RobertBColton: It's built into `qDebug`, which `qWarning` macro uses.
https://doc.qt.io/qt-5/qdebug.html#basic-use

Components/Logger.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <QDebug>
+#include <QtGlobal>
+
+class LogAndReturn {
+  QDebug logger;
+
+ public:
+  LogAndReturn(QtMsgType msgType) : logger(msgType) {}
+  LogAndReturn(const QDebug& log) : logger(log) {}
```
JoshDreamland: QDebug may implement move semantics (ie, you may be able to construct using `QDebug&&` from that `qWarning()` object.

RobertBColton: Ok: d99691125dcb9f17107fdca705b98c43cd4192fd

Qt seems cool with that.

RobertBColton: Also here's the definition of `qWarning()`.

```cpp
#define qWarning QMessageLogger(QT_MESSAGELOG_FILE, QT_MESSAGELOG_LINE, QT_MESSAGELOG_FUNC).warning
```

PR #65, opened 2019-05-24
New Include Icon

RobertBColton:
This is a friend of the same change in LateralGM. Josh has designed a new icon to use for included data files and we are using the old one for macros/definitions instead.
https://github.com/IsmAvatar/LateralGM/pull/436


RobertBColton:
I am not seeing what you are seeing, are looking at this from the command line? I don't think it will matter after I squash merge Josh. On GitHub it looks like a modification and addition.

JoshDreamland:
Binary files can't show as a modification. Assuming you meant move, go ahead. Not a huge deal regardless.

As the project gets bigger, we should talk about working around InkScape's bullshit by rendering the whole SVG and cutting it into pieces manually. I can modify Calico's raster script to export a list of icon names and coordinates, instead.

RobertBColton:
Exactly, but I will need you to approve it heh.

PR #66, opened 2019-06-14
 Change so resources are created in the default corresponding folder / Fixed room editor segfault due to pixmap.

polygoned:
2 commits:

- Change so resources are created in the default corresponding folder.
- Fixed room editor segfault due to pixmap.


MainWindow.h
```diff
@@ -99,7 +100,36 @@ class MainWindow : public QMainWindow {
...
+    TypeCase::kPath,
+    TypeCase::kScript,
+    TypeCase::kShader,
+    TypeCase::kFont,
+    TypeCase::kTimeline,
+    TypeCase::kObject,
+    TypeCase::kRoom,
+    TypeCase::kInclude,
+    TypeCase::kSettings
+  };
```
RobertBColton: This is redundant and already exists. Also it's noisy, so condense it.
https://github.com/enigma-dev/RadialGM/blob/fa091deac99ceae4fb12dcd74f193cdeb31dcb62/MainWindow.cpp#L300

PR #70, opened 2019-06-17
Typed Resource Groups and Crash Fixes

polygoned:
Pull for resource based system and some segfault fixes.


MainWindow.ui
```diff
@@ -823,7 +844,7 @@
      <normaloff>:/resources/timeline.png</normaloff>:/resources/timeline.png</iconset>
    </property>
    <property name="text">
-    <string>Create &amp;Timeline</string>
+    <string>Create Time Line</string>
```
RobertBColton: Undo this one, it's Timeline in ENIGMA. This is an ENIGMA IDE, not a GameMaker one. Please learn to spell properly.

MainWindow.cpp
```diff
@@ -316,8 +330,25 @@ void MainWindow::openProject(std::unique_ptr<buffers::Project> openedProject) {
...
   treeModel->connect(treeModel.get(), &TreeModel::ResourceRenamed, resourceMap.get(),
                      &ResourceModelMap::ResourceRenamed);
+
+  nodeResource.clear();
+  const QTreeView *treeUi = ui->treeView;
+  treeModel.get();
+  QModelIndex node = treeUi->indexAt(QPoint());
+  QVariant nodeString;
+  int index;
+  while (node.isValid()) {
```
RobertBColton: These are all things that you should be defensively programming and logging using `R_EXPECT` or `R_ASSESS`.

MainWindow.cpp
```diff
@@ -85,6 +87,14 @@ void diagnosticHandler(QtMsgType type, const QMessageLogContext &context, const
   }
 }
 
+void MainWindow::nameMap(const char *name, TypeCase type)
+{
+   QVariant nameVar = tr(name);
+//   defaultGroupNames.emplace_back(nameVar, type);
```
RobertBColton: Don't leave commented code in the source, it just rots. If it's not needed, delete it.

main.cpp
```diff
@@ -11,6 +11,9 @@
 QString defaultStyle = "";
 
 int main(int argc, char *argv[]) {
+  if (true) {
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);  //For correct high Dpi Scaling
+  }
```
RobertBColton: Remove the condition. Also consider making this an actual preference.

Widgets/RoomView.cpp
```diff
@@ -118,8 +126,9 @@ void RoomView::paintInstances(QPainter& painter, Room* room) {
     int yoff = 0;
 
     const ProtoModel* spr = GetObjectSprite(inst.object_type());
-    if (spr == nullptr)
+    if (spr == nullptr) {
       imgFile = "object";
+    }
     else {
```
RobertBColton: Either remove those braces you added or put the else on the same line. Either will be logically equivalent so it doesn't matter, but please do something other than what you have.

Widgets/RoomView.cpp
```diff
@@ -66,14 +70,18 @@ void RoomView::paintBackgrounds(QPainter& painter, Room* room, bool foregrounds)
...
     if (!bkgRes) continue;
     bkgRes = bkgRes->GetSubModel(TreeNode::kBackgroundFieldNumber);
     if (!bkgRes) continue;
     int w = bkgRes->data(Background::kWidthFieldNumber).toInt();
     int h = bkgRes->data(Background::kHeightFieldNumber).toInt();
 
     QString imgFile = bkgRes->data(Background::kImageFieldNumber).toString();
-    QPixmap pixmap = ArtManager::GetCachedPixmap(imgFile);
+    QPixmap pixmap;
+    ArtManager::GetCachedPixmap(imgFile, pixmap);
```
RobertBColton: Undo this, wth.

Widgets/RoomView.cpp
```diff
@@ -66,14 +70,18 @@ void RoomView::paintBackgrounds(QPainter& painter, Room* room, bool foregrounds)
   for (auto bkg : room->backgrounds()) {  //TODO: need to draw last if foreground
     if (!bkg.visible() || bkg.foreground() != foregrounds) continue;
     ProtoModel* bkgRes = MainWindow::resourceMap->GetResourceByName(TreeNode::kBackground, bkg.background_name());
+    if (bkgRes == nullptr) {
+      continue;
+    }
```
RobertBColton: This one is fine, but put it on a single line. Do NOT defensively log this one, since it's not an editor problem, it's _valid_ for the project to be in this state.
`if (bkgRes == nullptr) continue;`

Widgets/RoomView.cpp
```diff
@@ -50,7 +50,11 @@ void RoomView::paintTiles(QPainter& painter, Room* room) {
     int h = static_cast<int>(tile.height());
 
     QString imgFile = bkg->data(Background::kImageFieldNumber).toString();
-    QPixmap pixmap = ArtManager::GetCachedPixmap(imgFile);
+    if (bkg == nullptr) {
+      continue;
+    }
```
RobertBColton: Same here, put this one on a single line.
`if (bkg == nullptr) continue;`

RadialGM.pro.shared
```diff
@@ -44,4 +44,4 @@
             <value type="bool" key="EditorConfiguration.inEntireDocument">true</value>
         </valuemap>
     </data>
-</qtcreator>
+</qtcreator>
```
RobertBColton: You forget the end of file newline. Please add it, this is a standard convention originating from Unix.

MainWindow.cpp
```diff
@@ -29,10 +29,12 @@
 
 #include <functional>
 #include <unordered_map>
+#include <QtGlobal>
```
RobertBColton: Yeah, no, don't include this one if you have to. We already came up with a general not to do this until you've hit like 10 or 15 Qt includes. Only at that point or when you absolutely have to do we allow including this monolithic header.

Models/TreeModel.cpp
```diff
@@ -308,6 +309,7 @@ buffers::TreeNode *TreeModel::duplicateNode(const buffers::TreeNode &node) {
 
 void TreeModel::removeNode(const QModelIndex &index) {
   if (!index.isValid()) return;
+  MainWindow::nodeResource.remove(index);
```
RobertBColton: I don't really like this coupling of the tree model to the main window. Please instead have the main window call this in a slot which it connects to the tree model removed signal. That way the tree model doesn't have to know anything about the main window and we can use it in other ways.

RobertBColton: Here's the signal: <https://doc.qt.io/qt-5/qabstractitemmodel.html#rowsRemoved>

MainWindow.cpp
```diff
@@ -433,11 +464,36 @@ void MainWindow::CreateResource(TypeCase typeCase) {
...
+      }
+  }
+  else if (nodeResource.value(currentIndex, TypeCase::TYPE_NOT_SET) != typeCase && nodeResource.value(currentIndex.parent(), TypeCase::TYPE_NOT_SET) != typeCase) {
+      QList<QPersistentModelIndex> keys = nodeResource.keys(typeCase);
+      if (!keys.isEmpty()) {
+          QModelIndex oldNode = keys.constLast();  // use the oldest node mapped to that TypeCase
+          if (oldNode.isValid()) {
+              currentIndex = oldNode;
+          }      }
+  }
```
RobertBColton: Put all of these elses on the same line as the closing braces.

MainWindow.cpp
```diff
@@ -295,13 +311,11 @@ void MainWindow::openNewProject() {
...
-  QList<QString> defaultGroups = {tr("Sprites"), tr("Sounds"),  tr("Backgrounds"), tr("Paths"),
-                                  tr("Scripts"), tr("Shaders"), tr("Fonts"),       tr("Timelines"),
-                                  tr("Objects"), tr("Rooms"),   tr("Includes"),    tr("Configs")};
-  for (auto groupName : defaultGroups) {
+
+  for (auto& groupName : indexVector) {
     auto *groupNode = root->add_child();
     groupNode->set_folder(true);
-    groupNode->set_name(groupName.toStdString());
+    groupNode->set_name(groupName.name.toString().toStdString());
```
RobertBColton: You maybe should not need the intermediate `toString()`?

PR #72, opened 2019-06-18
Typed Resource Groups and Crash Fixes

polygoned:


Editors/RoomEditor.cpp
```diff
@@ -139,3 +139,9 @@ void RoomEditor::on_actionZoomIn_triggered() { ui->roomView->SetZoom(ui->roomVie
 void RoomEditor::on_actionZoomOut_triggered() { ui->roomView->SetZoom(ui->roomView->GetZoom() / 2); }
 
 void RoomEditor::on_actionZoom_triggered() { ui->roomView->SetZoom(1.0); }
+
+void RoomEditor::on_actionSave_triggered()
+{
+    QWidget *parent = parentWidget();
+    parent->close();  //pass just save in future
+}
```
RobertBColton: Since save is so common, there should be `defaultSaveAction` on `BaseEditor` or whatever, like LateralGM does. If an editor wants to override it or "augment" the default save behavior, they still can. There's just no reason to duplicate this code though.

main.cpp
```diff
@@ -11,6 +11,7 @@
 QString defaultStyle = "";
 
 int main(int argc, char *argv[]) {
+  QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);  //For correct high Dpi Scaling
```
RobertBColton: This has already been merged in #73 and #74 so no need.

Widgets/RoomView.cpp
```diff
@@ -66,14 +68,16 @@ void RoomView::paintBackgrounds(QPainter& painter, Room* room, bool foregrounds)
   for (auto bkg : room->backgrounds()) {  //TODO: need to draw last if foreground
     if (!bkg.visible() || bkg.foreground() != foregrounds) continue;
     ProtoModel* bkgRes = MainWindow::resourceMap->GetResourceByName(TreeNode::kBackground, bkg.background_name());
+    if (bkgRes == nullptr)  continue;
     if (!bkgRes) continue;
```
RobertBColton: The added `nullptr` check is redundant.

Components/ArtManager.cpp
```diff
@@ -29,13 +29,14 @@ const QIcon& ArtManager::GetIcon(const QString& name) {
...
   if (!QPixmapCache::find(name, &pm)) {
-    pm.load(name);
-    QPixmapCache::insert(name, pm);
+    if (!pm.load(name)) return false;
+     QPixmapCache::insert(name, pm);
   }
-  return std::move(pm);
+  return true;
+  //  return std::move(pm);
 }
```
RobertBColton: We found the actual reason this was a problem in #75 which fixed the issue without changing the callers.

Components/ArtManager.cpp
```diff
@@ -29,13 +29,14 @@ const QIcon& ArtManager::GetIcon(const QString& name) {
...
 
-const QPixmap& ArtManager::GetCachedPixmap(const QString& name) {
-  QPixmap pm;
+bool ArtManager::GetCachedPixmap(const QString& name, QPixmap &pm) {
+//  QPixmap pm;
   if (!QPixmapCache::find(name, &pm)) {
-    pm.load(name);
-    QPixmapCache::insert(name, pm);
+    if (!pm.load(name)) return false;
+     QPixmapCache::insert(name, pm);
```
JoshDreamland: `return QPixmapCache::insert(name, pm);`

This line might refuse to add a pixmap to the cache and return false.

MainWindow.cpp
```diff
@@ -570,3 +616,29 @@ void MainWindow::on_actionSortByName_triggered() {
...
+
+void MainWindow::ChangeNodeType(TypeCase type) {
+    if (!ui->treeView->selectionModel()->hasSelection()) return;
+    if (!nodeResource.contains(ui->treeView->currentIndex())) {
+      QPersistentModelIndex(ui->treeView->currentIndex());
+    }
+    nodeResource.insert(ui->treeView->currentIndex(), type);
+}
+
+void MainWindow::on_actionChange_to_Sprite_triggered() { ChangeNodeType(TypeCase::kSprite); }
```
JoshDreamland: Honestly, I'd just inline these in the class definition.

RobertBColton:
Closing this one out as stale. Things have evolved too much, and GMS2 itself no longer restricts the tree. People just don't want to work with restricted trees anymore, it doesn't provide any benefit and just increases UI clutter. If people really wanted this the correct way is to make a virtual view as noted above or move this functionality to a plugin. It's also not sensible with the new serialization protocol allowing potentially infinite resource types and prefabs.

PR #76, opened 2019-11-22
ProtoModel improvements

fundies:
fix crash in editors, enable save/close in editors, disable expanding instance to prevent crash on dbl click in room editor, force the room editor to constantly redraw


Models/RepeatedProtoModel.cpp
```diff
@@ -70,3 +73,11 @@ Qt::ItemFlags RepeatedProtoModel::flags(const QModelIndex &index) const {
   if (!index.isValid()) return nullptr;
   return QAbstractItemModel::flags(index);
 }
+
+bool RepeatedProtoModel::removeRows(int row, int count, const QModelIndex &parent) {
```
JoshDreamland: As discussed on Discord, this implementation would be quadratic and therefore choke when removing even a thousand elements. You need to implement `beginRemoveRows` and `endRemoveRows`, as well. Qt normally expects theses to be called automatically, but we'll need to call them on the entire container before/after chopping it up. Qt does not require these methods to be implemented at all, so feel free to rename.

```cpp
struct RowRemovalOperation() {
  int left, right;
  MutableRepeatedFieldRef<Message> *field;
  bool active = false;
} rmRows_;

void RepeatedProtoModel::beginRemoveRows(const QModelIndex &parent, int first, int last) {
  rmRows_.left = rmRows_.right = 0;
  const Reflection *refl = protobuf->GetReflection();
  rmRows_.field = refl->GetMutableRepeatedFieldRef<Message>(protobuf, field);
  rmRows_.active = true;
  emit rowsAboutToBeRemoved(parent, first, last);
}

bool RepeatedProtoModel::removeRows(int row, int count, const QModelIndex &parent) {
  if (!rmRows_.active) return false;
  if (rmRows_.left < rmRows_.right) while (rmRows_.right < row) {
    rmRows_.field.SwapElements(rmRows_.left++, rmRows_.right++);
  }
  rmRows_.right = row + count;
  return true;
}

void RepeatedProtoModel::endRemoveRows() {
  if (!rmRows_.active) return;
  rmRows_.active = false;
  if (rmRows_.left < rmRows_.right) {
    while (rmRows_.right < mutableRef.size()) {
      rmRows_.field.SwapElements(rmRows_.left++, rmRows_.right++);
    }
    mutableRef.Truncate(rmRows_.left);
    emit dataChanged(index(0), index(rmRows_.left));
  }
}
```

PR #80, opened 2019-12-23
Fonts

fundies:


Editors/FontEditor.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>369</width>
-    <height>448</height>
+    <width>773</width>
+    <height>443</height>
```
JoshDreamland: lol... I'm so glad we have WYSIWYG for this shit

Models/RepeatedProtoModel.cpp
```diff
@@ -105,7 +109,7 @@ QModelIndex RepeatedProtoModel::index(int row, int column, const QModelIndex & /
 
 Qt::ItemFlags RepeatedProtoModel::flags(const QModelIndex &index) const {
   if (!index.isValid()) return nullptr;
-  return QAbstractItemModel::flags(index);
+  return QAbstractItemModel::flags(index) | Qt::ItemIsEditable;
```
JoshDreamland: Are we supposed to do this manually instead of telling our model it's editable? Do we always want these to be?

fundies: It's easiest to just set the checkbox in the view to say its not editable

PR #81, opened 2019-12-25
Fix Room Editor CPU Throttling

RobertBColton:
Calling update in the paint event is a sin.

https://doc.qt.io/qt-5/qwidget.html#updatesEnabled-prop
https://doc.qt.io/qt-5/qwidget.html#update

### DIFF
--- a/Widgets/RoomView.cpp
+++ b/Widgets/RoomView.cpp
@@ -64,7 +64,7 @@ void RoomView::paintEvent(QPaintEvent* /* event */) {
   this->paintBackgrounds(painter, true);
   this->paintGrid(painter);
 
-  update();
+  //update();
 }
 
 void RoomView::paintTiles(QPainter& painter) {


fundies:
You can't just remove that line. Without it, the room won't properly draw on any changes. I've made steps to hook up a recursive signal that catches any changes and then calls update() but thats comming in my next PR. For now, removing this shim would completly break rooms.

JoshDreamland:
How about a simple ticker that just updates every 30 milliseconds? As-is, this thing devours CPU.

RobertBColton:
Then call it from a timer thread at 60 fps, otherwise this is just recursive on the main thread. It's not calling update that's the problem, it's how its being called.

The room doesn't really need repainted at 60fps though, even for animated sprites you can just clear a subrect and only refresh those sprites. Default behavior is repaint when resized.

fundies:
Just wait for my next PR. It will only update when there are changes to any model when I'm done. I got the framework mostly there in my last pr but I'll hook up rest in this upcomimming one.

RobertBColton:
Good enough.

PR #82, opened 2020-01-05
buggy paths

fundies:
This PR adds support for paths, refactors asset drawing to allow drawing out of bounds, expands the grid in room editor to the whole window, changes room to only redraw on model updates, and probably 3 other things I forgot.

### DIFF
--- a/Editors/BackgroundEditor.cpp
+++ b/Editors/BackgroundEditor.cpp
@@ -45,15 +45,11 @@ void BackgroundEditor::dataChanged(const QModelIndex& topLeft, const QModelIndex
   ui->backgroundView->update();
 }
 
-void BackgroundEditor::on_actionZoomIn_triggered() {
-  ui->imagePreviewBackground->SetZoom(ui->imagePreviewBackground->GetZoom() * 2);
-}
+void BackgroundEditor::on_actionZoomIn_triggered() { ui->imagePreviewBackground->ZoomIn(); }
 
-void BackgroundEditor::on_actionZoomOut_triggered() {
-  ui->imagePreviewBackground->SetZoom(ui->imagePreviewBackground->GetZoom() / 2);
-}
+void BackgroundEditor::on_actionZoomOut_triggered() { ui->imagePreviewBackground->ZoomOut(); }
 
-void BackgroundEditor::on_actionZoom_triggered() { ui->imagePreviewBackground->SetZoom(1); }
+void BackgroundEditor::on_actionZoom_triggered() { ui->imagePreviewBackground->ResetZoom(); }
 
 void BackgroundEditor::on_actionNewImage_triggered() {
   QDialog dialog(this);
@@ -82,8 +78,9 @@ void BackgroundEditor::on_actionLoadImage_triggered() {
         // QString newData = GetModelData(Background::kImageFieldNumber).toString();
         // TODO: Copy data into our egm and reset the path
         // SetModelData(Background::kImageFieldNumber, lastData);
-      } else
+      } else {
         qDebug() << "Failed to load gmx Background";
+      }
     } else {
       // TODO: Copy data into our egm
       SetModelData(Background::kImageFieldNumber, fName);
--- a/Editors/BackgroundEditor.cpp
+++ b/Editors/BackgroundEditor.cpp
@@ -9,9 +9,9 @@
 
 #include "gmx.h"
 
+#include <QDebug>
 #include <QDesktopServices>
 #include <QPainter>
-#include <QDebug>
 
 using buffers::resources::Background;
 
@@ -20,6 +20,8 @@ BackgroundEditor::BackgroundEditor(ProtoModelPtr model, QWidget* parent)
   ui->setupUi(this);
 
   connect(ui->actionSave, &QAction::triggered, this, &BaseEditor::OnSave);
+
+  ui->imagePreviewBackground->SetAssetView(ui->backgroundView);
 
   ui->backgroundView->SetResourceModel(resMapper->GetModel());
 
@@ -44,14 +46,14 @@ void BackgroundEditor::dataChanged(const QModelIndex& topLeft, const QModelIndex
 }
 
 void BackgroundEditor::on_actionZoomIn_triggered() {
-  ui->backgroundView->SetZoom(ui->backgroundView->GetZoom() * 2);
+  ui->imagePreviewBackground->SetZoom(ui->imagePreviewBackground->GetZoom() * 2);
 }
 
 void BackgroundEditor::on_actionZoomOut_triggered() {
-  ui->backgroundView->SetZoom(ui->backgroundView->GetZoom() / 2);
+  ui->imagePreviewBackground->SetZoom(ui->imagePreviewBackground->GetZoom() / 2);
 }
 
-void BackgroundEditor::on_actionZoom_triggered() { ui->backgroundView->SetZoom(1); }
+void BackgroundEditor::on_actionZoom_triggered() { ui->imagePreviewBackground->SetZoom(1); }
 
 void BackgroundEditor::on_actionNewImage_triggered() {
   QDialog dialog(this);
@@ -80,7 +82,8 @@ void BackgroundEditor::on_actionLoadImage_triggered() {
         // QString newData = GetModelData(Background::kImageFieldNumber).toString();
         // TODO: Copy data into our egm and reset the path
         // SetModelData(Background::kImageFieldNumber, lastData);
-      } else qDebug() << "Failed to load gmx Background";
+      } else
+        qDebug() << "Failed to load gmx Background";
     } else {
       // TODO: Copy data into our egm
       SetModelData(Background::kImageFieldNumber, fName);
--- a/Editors/PathEditor.cpp
+++ b/Editors/PathEditor.cpp
@@ -146,12 +146,13 @@ void PathEditor::RebindSubModels() {
     if (button == Qt::MouseButton::LeftButton) {
       draggingPoint = true;
       // Check for point at location (reverse loop to get last added one if dups)
-      for (int i = pointsModel->rowCount(); i >= 0; i--) {
+      for (int i = pointsModel->rowCount() - 1; i >= 0; i--) {
         QPoint pt(pointsModel->data(i, Path::Point::kXFieldNumber).toInt(),
                   pointsModel->data(i, Path::Point::kYFieldNumber).toInt());
         if (pt == ui->roomView->mousePos) {
-          ui->roomView->selectedPointIndex = i;
-          ui->pointsTableView->selectionModel()->select(pointsModel->index(i, Path::Point::kXFieldNumber),
+          QModelIndex newSelectIndex = pointsModel->index(i, Path::Point::kXFieldNumber);
+          ui->pointsTableView->setCurrentIndex(newSelectIndex);
+          ui->pointsTableView->selectionModel()->select(newSelectIndex,
                                                         QItemSelectionModel::QItemSelectionModel::ClearAndSelect);
           return;
         }
@@ -183,12 +184,14 @@ void PathEditor::RebindSubModels() {
 }
 
 bool PathEditor::eventFilter(QObject* obj, QEvent* event) {
-  // Resize columns to view size
-  if (obj == ui->pointsTableView && event->type() == QEvent::Resize) {
-    QResizeEvent* resizeEvent = static_cast<QResizeEvent*>(event);
-    for (int c = 0; c < 3; ++c) {
-      // Note: I ha nfi what 40 is here. Something todo with margins prolly
-      ui->pointsTableView->setColumnWidth(c, (resizeEvent->size().width() - 40) / 3);
+  if (pointsModel != nullptr) {
+    // Resize columns to view size
+    if (obj == ui->pointsTableView && event->type() == QEvent::Resize) {
+      QResizeEvent* resizeEvent = static_cast<QResizeEvent*>(event);
+      int cc = pointsModel->columnCount() - 1;
+      for (int c = 1; c < cc; ++c) {  // column 1 is hidden
+        ui->pointsTableView->setColumnWidth(c, (resizeEvent->size().width()) / cc);
+      }
     }
   }
   return QWidget::eventFilter(obj, event);
@@ -219,12 +222,15 @@ void PathEditor::on_deletePointButton_pressed() {
     remover.RemoveRow(deleteIndex);
   }
 
-  qDebug() << deleteIndex;
-
-  if (pointsModel->rowCount() > 0)
-    ui->pointsTableView->selectionModel()->select(
-        pointsModel->index((deleteIndex == 0) ? 0 : deleteIndex - 1, Path::Point::kXFieldNumber),
-        QItemSelectionModel::QItemSelectionModel::ClearAndSelect);
-  else
+  if (pointsModel->rowCount() > 0) {
+    QModelIndex newSelectIndex =
+        pointsModel->index((deleteIndex == 0) ? 0 : deleteIndex - 1, Path::Point::kXFieldNumber);
+    ui->pointsTableView->setCurrentIndex(newSelectIndex);
+    ui->pointsTableView->selectionModel()->select(newSelectIndex,
+                                                  QItemSelectionModel::QItemSelectionModel::ClearAndSelect);
+  } else {
     ui->deletePointButton->setDisabled(true);
+    ui->pointsTableView->selectionModel()->clearSelection();
+    ui->pathPreviewBackground->update();
+  }
 }
--- a/Editors/PathEditor.cpp
+++ b/Editors/PathEditor.cpp
@@ -87,14 +87,7 @@ PathEditor::PathEditor(ProtoModelPtr model, QWidget* parent) : BaseEditor(model,
   connect(ui->actionSnaptoGrid, &QAction::triggered, [this]() { this->snapToGrid = !this->snapToGrid; });
 
   cursorPositionLabel = new QLabel();
-  connect(ui->pathPreviewBackground, &AssetScrollAreaBackground::MouseMoved, [=](int x, int y) {
-    const GridDimensions g = ui->roomView->GetGrid();
-    QPoint mousePos(x, y);
-    if (snapToGrid) mousePos = QPoint(RoundNum(x, g.horSpacing), RoundNum(y, g.vertSpacing));
-    ui->roomView->mousePos = mousePos;
-    cursorPositionLabel->setText(tr("X %0, Y %1").arg(mousePos.x()).arg(mousePos.y()));
-    ui->pathPreviewBackground->update();
-  });
+
   ui->statusBar->addWidget(cursorPositionLabel);
 
   ui->pointsTableView->installEventFilter(this);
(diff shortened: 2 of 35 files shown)


fundies:
@JoshDreamland this ready for review now

fundies:
This does crash seemingly randomly when I spam the delete button sometimes with qvector out of range error but I cant reproduce it regularly

Editors/PathEditor.cpp
```diff
@@ -1,10 +1,238 @@
...
+    // Resize columns to view size
+    if (obj == ui->pointsTableView && event->type() == QEvent::Resize) {
+      QResizeEvent* resizeEvent = static_cast<QResizeEvent*>(event);
+      int cc = pointsModel->columnCount() - 1;
+      for (int c = 1; c < cc; ++c) {  // column 1 is hidden
+        ui->pointsTableView->setColumnWidth(c, (resizeEvent->size().width()) / cc);
+      }
+    }
+  }
+  return QWidget::eventFilter(obj, event);
```
JoshDreamland: Not familiar enough with Qt to know if this is always supposed to be called. You're handling some events above. Is this what invokes the base view's event handler?

Editors/PathEditor.cpp
```diff
@@ -1,10 +1,238 @@
...
+
+  QToolButton* zoom(new QToolButton(this));
+  zoom->setIcon(ArtManager::GetIcon(":/actions/zoom.png"));
+  ui->mainToolBar->addWidget(zoom);
+  connect(zoom, &QToolButton::pressed, ui->pathPreviewBackground, [=]() { ui->pathPreviewBackground->SetZoom(1); });
+
+  QToolButton* zoomIn(new QToolButton(this));
+  zoomIn->setIcon(ArtManager::GetIcon(":/actions/zoom-in.png"));
+  ui->mainToolBar->addWidget(zoomIn);
+  connect(zoomIn, &QToolButton::pressed, ui->pathPreviewBackground,
```
JoshDreamland: Shouldn't the background object just expose a function to handle zoom, instead of each editor hard-coding a default zoom factor?

Editors/PathEditor.cpp
```diff
@@ -1,10 +1,238 @@
...
+      ui->pathPreviewBackground->update();  // manually call update to redraw cursor if no data changes
+  });
+
+  connect(ui->pathPreviewBackground, &AssetScrollAreaBackground::MousePressed, [=](Qt::MouseButton button) {
+    if (button == Qt::MouseButton::LeftButton) {
+      draggingPoint = true;
+      // Check for point at location (reverse loop to get last added one if dups)
+      for (int i = pointsModel->rowCount() - 1; i >= 0; i--) {
+        QPoint pt(pointsModel->data(i, Path::Point::kXFieldNumber).toInt(),
+                  pointsModel->data(i, Path::Point::kYFieldNumber).toInt());
```
JoshDreamland: There's a Point() accessor on PathView for this, now.

Widgets/AssetScrollAreaBackground.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <QSet>
+#include <QWidget>
+
+#include <cmath>
+
+inline bool compareDouble(double a, double b) { return std::fabs(a - b) < std::numeric_limits<double>::epsilon(); }
+
+// RoundNum round to the nearest multiple
+template <typename T>
```
JoshDreamland: This probably belongs in a utility header; here, we're just rounding points, so why not just accept a QPoint?

Widgets/AssetView.h
```diff
@@ -1,23 +1,41 @@
...
+
+struct GridDimensions {
+  GridType type = GridType::Standard;
+  bool show = true;
+  int horSpacing = 16;
+  int vertSpacing = 16;
+  int horOff = 0;
+  int vertOff = 0;
+  int cellWidth = 0;
+  int cellHeight = 0;
```
JoshDreamland: Another idea is to define this as a unitCell rectangle, capturing horSpacing, vertSpacing, cellWidth, and cellHeight. The grid is then simply a tiled matrix of unit cells.

Editors/BackgroundEditor.cpp
```diff
@@ -80,7 +82,8 @@ void BackgroundEditor::on_actionLoadImage_triggered() {
         // QString newData = GetModelData(Background::kImageFieldNumber).toString();
         // TODO: Copy data into our egm and reset the path
         // SetModelData(Background::kImageFieldNumber, lastData);
-      } else qDebug() << "Failed to load gmx Background";
+      } else
+        qDebug() << "Failed to load gmx Background";
```
JoshDreamland: clang-format should have put braces around this; what's up

Editors/PathEditor.cpp
```diff
@@ -1,10 +1,238 @@
...
+  resMapper->addMapping(xSnap, Path::kHsnapFieldNumber);
+  resMapper->addMapping(ySnap, Path::kVsnapFieldNumber);
+
+  RebindSubModels();
 }
 
 PathEditor::~PathEditor() { delete ui; }
+
+void PathEditor::RebindSubModels() {
+  pathModel = _model->GetSubModel(TreeNode::kPathFieldNumber);
```
JoshDreamland: What prevents this from going stale?

Widgets/AssetScrollAreaBackground.h
```diff
@@ -0,0 +1,60 @@
...
+  void paintEvent(QPaintEvent* event) override;
+  bool eventFilter(QObject* obj, QEvent* event) override;
+
+  AssetView* assetView = nullptr;
+  bool drawSolidBackground = true;
+  qreal zoom = 1;
+  QPoint offset;
+  QPoint userOffset;
+  QSet<int> pressedKeys;
+  QColor backgroundColor = Qt::GlobalColor::gray;
```
JoshDreamland: I had wished a couple of these ended in `_` so I'd know it was a local.

Widgets/BackgroundView.cpp
```diff
@@ -54,27 +56,28 @@ void BackgroundView::WriteImage(QString fName, QString type) {
...
-  int gridVertOff = model->data(Background::kVerticalOffsetFieldNumber).toInt();
-  int gridWidth = model->data(Background::kTileWidthFieldNumber).toInt();
-  int gridHeight = model->data(Background::kTileHeightFieldNumber).toInt();
-
-  if (gridVisible) {
-    paintGrid(painter, pixmap.width(), pixmap.height(), gridHorSpacing, gridVertSpacing, gridHorOff, gridVertOff,
-              gridWidth, gridHeight);
+  if (model->data(Background::kUseAsTilesetFieldNumber).toBool()) {
+    grid.show = true;
+    grid.horSpacing = model->data(Background::kHorizontalSpacingFieldNumber).toInt();
```
JoshDreamland: In the future, it might be cool to allow binding the ScrollAreaBackground to these fields, directly. Especially if we add a UI-specific field to paths for persisting grid settings, as we've done for rooms.

Incidentally, maybe we should consider extracting those into a file we can `.gitignore`, eg, `roomname.rm.ide`.

Widgets/RoomView.cpp
```diff
@@ -22,49 +22,55 @@ bool InstanceSortFilterProxyModel::lessThan(const QModelIndex &left, const QMode
...
+  repaint();
 }
 
 QSize RoomView::sizeHint() const {
   if (!model) return QSize(640, 480);
-  unsigned roomWidth = model->data(Room::kWidthFieldNumber).toUInt(),
-           roomHeight = model->data(Room::kHeightFieldNumber).toUInt();
-  return QSize(roomWidth, roomHeight);
+  QVariant roomWidth = model->data(Room::kWidthFieldNumber), roomHeight = model->data(Room::kHeightFieldNumber);
+  if (!roomWidth.isValid() || !roomHeight.isValid()) return QSize(640, 480);
```
JoshDreamland: Maybe flag this with a TODO(enigma-dev/enigma-dev#1872)

Widgets/RoomView.cpp
```diff
@@ -22,49 +22,55 @@ bool InstanceSortFilterProxyModel::lessThan(const QModelIndex &left, const QMode
...
 
 void RoomView::SetResourceModel(ProtoModelPtr model) {
   this->model = model;
-  this->sortedInstances->setSourceModel(model->GetRepeatedSubModel(Room::kInstancesFieldNumber));
-  this->sortedInstances->sort(Room::Instance::kObjectTypeFieldNumber);
-  this->sortedTiles->setSourceModel(model->GetRepeatedSubModel(Room::kTilesFieldNumber));
-  this->sortedTiles->sort(Room::Tile::kDepthFieldNumber);
-  this->SetZoom(1.0);
+
+  if (model != nullptr) {
```
JoshDreamland: Out of curiosity, what prompted you to add this check?

Widgets/ResourceSelector.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "Components/QMenuView.h"
+#include "MainWindow.h"
+#include "Models/TreeSortFilterProxyModel.h"
+
+#include <QToolButton>
+
+class ResourceSelector : public QToolButton {
+ public:
+  explicit ResourceSelector(QWidget* parent, TreeNode::TypeCase type = TreeNode::TypeCase::kObject)
```
JoshDreamland: I feel dirty defaulting this to object. This is just so this constructor can serve as the default, yes?

Editors/BackgroundEditor.cpp
fundies: If you know the clang format option to fix it I'll add it but I don't see one

Editors/PathEditor.cpp
fundies: I didn't think it mattered for a 1 loc feature but I can probably inherit it I guess

Editors/PathEditor.cpp
fundies: The base model (or treenode) is only explicitly deleted by the user. If _model is null the editor should be closed

Editors/PathEditor.cpp
fundies: Ok whatever. I'll use it

Editors/PathEditor.cpp
fundies: Event filters capture events from any qobject you give a pointer to with installeventfiltet(obj) I'm only capturing the tableviews events here so only those are handled here. Qt has no option to equally stretch columns it can only stretch the last so this is only way to keep even columns other than making our own tableviews inherited and installing an eventfilter there instead

fundies: Event filters capture events from any qobject you give a pointer to with installeventfiltet(obj) I'm only capturing the tableviews events here so only those are handled here. Qt has no option to equally stretch columns it can only stretch the last so this is only way to keep even columns other than making our own tableviews inherited and installing an eventfilter there instead

MainWindow.cpp
```diff
@@ -154,6 +144,18 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui(new Ui::MainWi
   connect(ui->actionDebug, &QAction::triggered, pluginServer, &RGMPlugin::Debug);
   connect(ui->actionCreateExecutable, &QAction::triggered, pluginServer, &RGMPlugin::CreateExecutable);
 
+  connect(ui->mdiArea, &QMdiArea::subWindowActivated, [=](QMdiSubWindow *w) {
+    for (QMdiSubWindow *subWindow : subWindows) {
+      if (w == nullptr) continue;
```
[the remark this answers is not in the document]
fundies: When w is nullptr it means there is no active mdi window. The w there should actually be subwindow as we want to emit focus lost on subwindow if there is for some reason no active subwindow. I'll correct that

Models/RepeatedProtoModel.cpp
```diff
@@ -159,6 +160,14 @@ bool RepeatedProtoModel::insertRows(int row, int count, const QModelIndex &paren
   return true;
 }
 
+void RepeatedProtoModel::RecursiveDataChanged() {
```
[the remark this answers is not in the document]
fundies: Ok transvestite datachanged or whatever it is

Models/RepeatedProtoModel.h
```diff
@@ -56,10 +56,8 @@ class RepeatedProtoModel : public QAbstractItemModel {
     std::set<int> rows;
 
    public:
-    RowRemovalOperation(RepeatedProtoModelPtr m):
-        model(m),
-        field(m->protobuf->GetReflection()
-               ->GetMutableRepeatedFieldRef<Message>(m->protobuf, m->field)) {}
+    RowRemovalOperation(RepeatedProtoModelPtr m)
+        : model(m), field(m->protobuf->GetReflection()->GetMutableRepeatedFieldRef<Message>(m->protobuf, m->field)) {}
```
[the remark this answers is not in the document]
fundies: So?

Widgets/AssetScrollArea.h
```diff
@@ -0,0 +1,12 @@
+#ifndef ASSETPREVIEWSCROLLAREA_H
+#define ASSETPREVIEWSCROLLAREA_H
+
+#include <QScrollArea>
+
+class AssetScrollArea : public QScrollArea {
```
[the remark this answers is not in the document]
fundies: What do you mean? You need explicit paint on middle click? Why?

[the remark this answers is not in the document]
fundies: What do you mean? You need explicit paint on middle click? Why?

Widgets/AssetScrollAreaBackground.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    for (int y = gridVertOff; y > 0; y -= gridVertSpacing) {
+      painter.drawLine(0, y, this->width(), y);
+    }
+  }
+
+  painter.restore();
+}
+
+void AssetScrollAreaBackground::paintGrid(QPainter& painter, int width, int height, int gridHorSpacing,
```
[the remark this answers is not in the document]
fundies: That's the plan yes. But really any visual asset inherits this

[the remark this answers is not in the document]
fundies: That's the plan yes. But really any visual asset inherits this

Widgets/AssetScrollAreaBackground.h
fundies: I just stole this from a previous project. I dunno where else we would need this and we don't really have a utility header atm as qt does most our utility for us. I'll move it if and when I need it somewhere else. As for accepting qpoint I think it'd make for an ugly signature needing two rounder values accepted for each point. I think it's clearer what you're rounding x and y to this way

Widgets/AssetScrollAreaBackground.h
fundies: Well we aren't very consistent with this anywhere. I think sometime before release I'll try to add doxygen and clean this kinda stuff up too. But for now IDGAF. We can be more not picky on style when the thing works

Widgets/AssetView.h
fundies: I don't feel like breaking that function atm.

Widgets/BackgroundView.cpp
fundies: Ok I'll mark this with a to-do but I already do this in a few places. I wanted that issue resolved before this practice becomes too widespread so hopefully we can do something about either in or around the egm pr

Widgets/ResourceSelector.h
fundies: Without a default argument the designer's generated code craps out. I can default the type to none if you prefer but I do need a default

Widgets/RoomView.cpp
fundies: When you unset the room in the pathview you set it to nullptr so without this check it'll crash

fundies:
See above comment related to grid. It's the same issue

Widgets/BackgroundView.cpp
fundies: Paths already has this ui persistent option for grids

Widgets/PathView.cpp
```diff
@@ -0,0 +1,178 @@
...
+  RepeatedProtoModelPtr pointsModel = pathModel->GetRepeatedSubModel(Path::kPointsFieldNumber);
+  return QPoint(pointsModel->data(n, Path::Point::kXFieldNumber).toInt(),
+                pointsModel->data(n, Path::Point::kYFieldNumber).toInt());
+}
+
+namespace {
+
+// Draws a curve of adjustable shittiness. Use increment to control the
+// shittiness. For added shittiness, the start and end points will not be drawn.
+void shittyCurve(QPoint start, QPoint handle, QPoint end, double increment, QVector<QPoint> *points) {
```
fundies: you got better name for this yet?

Models/RepeatedProtoModel.cpp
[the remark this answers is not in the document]
fundies: whats the actual word for this?

Widgets/BackgroundView.cpp
fundies: Not exactly sure what you mean but paths do have a persistent grid. This sounds like a later task either way.

Models/RepeatedProtoModel.h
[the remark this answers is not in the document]
JoshDreamland: So it's taking code that was wrapped to 80-char limits (the norm) and re-wrapping it to something more like 200.

Models/RepeatedProtoModel.cpp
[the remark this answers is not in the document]
JoshDreamland: Iterative, but you probably just want to call it DataChangedThroughHierarchy or DataChangedForParents or something.

Editors/PathEditor.cpp
JoshDreamland: I see. So these are (or at least _model is) a weak pointer, and the editor will simply close when this would otherwise go stale. Sounds fine, although might lead to annoying behavior around, eg, undoing and redoing add or delete operations.

Editors/PathEditor.cpp
JoshDreamland: I should clarify: I get why we call it in general, but I'm not sure we should be calling it when we've already handled the event. But assuming you're seeing the right behavior for events we *are* handling here, this should be fine.

Widgets/RoomView.cpp
JoshDreamland: (I'd keep the issue links so they're easy to search, although we really lack the infrastructure for this to matter right now. It's possible GitHub may one day support it.)

Widgets/AssetScrollArea.h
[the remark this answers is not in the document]
JoshDreamland: I worry that's not discoverable. These editors are primarily mouse-driven.

Models/RepeatedProtoModel.h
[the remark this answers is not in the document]
fundies: I have 120 set. Robert and I felt 80 was too low

fundies:
> The way you implemented room rendering in the path editor is pretty cool. I will think about how to use this when we go to add Overworlds. I think we'll need to add a caching layer to it, so that when rendering at less than 25% scale, only the cached version is used.
> 
> Anyway, if you're done, I'll merge this.

I would like to rename your "shitty curve" functions can you come up with better names and I'' rename that plus my datachanged crap. then should be good to merge

JoshDreamland:
We can use "approximate curve" or something.

JoshDreamland:
So apparently control-enter closes and comments...?

PR #87, opened 2020-02-16
only 142 tries (so far)

fundies:
Good to go I think


azure-pipelines.yml
```diff
@@ -0,0 +1,113 @@
...
+    displayName: 'Debug Build'
+
+  - script: |
+      7z a $(Build.SourcesDirectory)/RadialGM-MSVC-x64.exe -mmt -mx5 -sfx $(Build.SourcesDirectory)/Submodules/enigma-dev
+    displayName: 'Package'
+
+  - task: GithubRelease@0 
+    displayName: 'Create GitHub Release'      
+    inputs:
+      gitHubConnection: github.com_fundies
```
fundies: need to update this once we setup azure

azure-pipelines.yml
```diff
@@ -0,0 +1,113 @@
...
+      repositoryName: fundies/RadialGM
+      tagSource: manual
+      tag: RadialGM-$(Date:yyyyMMdd)$(Rev:.r)
```
fundies: @RobertBColton any idea how I can make this match up with appveyor's tag? Cause my only idea is using date but I don't like that

RobertBColton: What's wrong with just $Rev? All we'd need is shared environment variable. If that's not doable Josh said before it's ok to check it into a text file.

RobertBColton:
You think Gandhi gave up after 142 tries? Well I got news for you, he gave up after 241 tries. You're only halfway there.

azure-pipelines.yml
fundies: and how is appveyor going to know what azxure's $rev var is?

RobertBColton: I don't know what $rev is, can you tell me? It used to be used in LGM's about dialog too.
https://github.com/IsmAvatar/LateralGM/commit/76e25f9d519ceaf1d800eee74f06e417b5924d79
I removed it assuming it was an artifact of SVN and replaced it with a static string of LGM's version number so I could update the version number easier and uniformly.

fundies:
@RobertBColton or @JoshDreamland you can review n merge this it appears to work now. The deploy is untested though cause its a PR. Also the msys build appear to be broken (they compile but segfault) but I think its just a grpc bug.

fundies:
@JoshDreamland so I got mingw working too so you can disable appveyor for this repo when you merge

PR #90, opened 2020-03-07
Restore GRPC

RobertBColton:
This is just the start of my GRPC fixes that I've been discussing. First, I am enabling the server plugin again. Then I fixed the file search pattern to whitelist "emake" and "emake.exe" so that "emake-tests.exe" is no longer matched. I went with this for now because I see Qt only accepts globs, not actual regexes, so the regex Josh gave me won't work yet. Next, I changed the timer delay to 1ms to prevent the plugin from hogging 100% of a CPU core. Finally, I added `rgm_enable_grpc_server` config option (disabled by default) in both the pro file and the CMakeLists to make it easier to turn the GRPC server on and off, and so I can keep working on it without undoing fundies commenting out my code.

### DIFF
--- a/MainWindow.cpp
+++ b/MainWindow.cpp
@@ -125,6 +125,7 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainW
   this->_recentFiles = new RecentFiles(this, this->_ui->menuRecent, this->_ui->actionClearRecentMenu);
 
   _ui->mdiArea->setBackground(QImage(":/banner.png"));
+  connect(_ui->mdiArea, &QMdiArea::subWindowActivated, this, &MainWindow::MDIWindowChanged);
   connect(_ui->menuWindow, &QMenu::aboutToShow, this, &MainWindow::updateWindowMenu);
 
   auto settingsButton = static_cast<QToolButton *>(_ui->mainToolBar->widgetForAction(_ui->actionSettings));
@@ -132,6 +133,7 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainW
   settingsButton->setToolButtonStyle(Qt::ToolButtonStyle::ToolButtonTextBesideIcon);
   _ui->actionSettings->setMenu(_ui->menuChangeGameSettings);
 
+#ifdef RGM_SERVER_ENABLED
   RGMPlugin *pluginServer = new ServerPlugin(*this);
   auto outputTextBrowser = this->_ui->outputTextBrowser;
   connect(pluginServer, &RGMPlugin::LogOutput, outputTextBrowser, &QTextBrowser::append);
@@ -145,8 +147,7 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainW
   connect(_ui->actionRun, &QAction::triggered, pluginServer, &RGMPlugin::Run);
   connect(_ui->actionDebug, &QAction::triggered, pluginServer, &RGMPlugin::Debug);
   connect(_ui->actionCreateExecutable, &QAction::triggered, pluginServer, &RGMPlugin::CreateExecutable);
-
-  connect(_ui->mdiArea, &QMdiArea::subWindowActivated, this, &MainWindow::MDIWindowChanged);
+#endif
 
   openNewProject();
 }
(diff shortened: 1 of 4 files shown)


RobertBColton:
Really I didn't need to do cc34f3d9d9fa6f6ede6a1af302463b2bd7ef1a5a because you were already indiscriminately building the server plugin into RGM (and really who's going to use it without the plugin) but I did anyway and it's analogous to the one I added to the pro file. In both cases it is off by default.

RobertBColton:
@fundies Shouldn't we honestly turn it on by default though? And you can just disable it until you get your shit sorted out?

Plugins/ServerPlugin.cpp
```diff
@@ -240,11 +240,6 @@ void CompilerClient::UpdateLoop() {
 }
 
 ServerPlugin::ServerPlugin(MainWindow& mainWindow) : RGMPlugin(mainWindow) {
-  /*
```
JoshDreamland: You don't seem to be checking the new macros you created. This code's just always uncommented. Are you just relying on it not being called?

MainWindow.cpp
```diff
@@ -125,13 +125,15 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainW
...
   _ui->mdiArea->setBackground(QImage(":/banner.png"));
+  connect(_ui->mdiArea, &QMdiArea::subWindowActivated, this, &MainWindow::MDIWindowChanged);
   connect(_ui->menuWindow, &QMenu::aboutToShow, this, &MainWindow::updateWindowMenu);
 
   auto settingsButton = static_cast<QToolButton *>(_ui->mainToolBar->widgetForAction(_ui->actionSettings));
   settingsButton->setPopupMode(QToolButton::ToolButtonPopupMode::MenuButtonPopup);
   settingsButton->setToolButtonStyle(Qt::ToolButtonStyle::ToolButtonTextBesideIcon);
   _ui->actionSettings->setMenu(_ui->menuChangeGameSettings);
 
+#ifdef RGM_SERVER_ENABLED
```
JoshDreamland: Since you're linking in all the code regardless of whether you intend to use it, this could as well be a runtime check.

Also, we're leaking that pluginServer ref. It should probably be a `unique_ptr` owned by this class (can be left null when disabled).

Plugins/ServerPlugin.cpp
RobertBColton: I am checking the define `RGM_SERVER_ENABLED` in `MainWindow.cpp` which calls this constructor. The plugin interface I designed was meant to be pluggable into RGM, so fundies kind of went commenting it out in the wrong place.

MainWindow.cpp
RobertBColton: Are you suggesting making it a dll? Yeah that's how Qt Creator has its own plugin interface, I just wasn't ready to set it up yet.

And no, we're not, Qt has its own ownership policy. The server plugin reference is owned by the main window. This is because the server plugin inherits from my generic RGM plugin abstraction which itself is based on QObject. All RGM plugins are informed of the main window at the present time.
https://github.com/enigma-dev/RadialGM/blob/6d626d725f3132f829d54028a5556abb38d89a09/Plugins/RGMPlugin.h#L31

JoshDreamland: No, I'm just suggesting being consistent about whether we're eliminating code at compile time or branching over it at runtime. My complaint above is related.

> I am checking the define `RGM_SERVER_ENABLED` in `MainWindow.cpp` which calls this constructor.

You are building an entire unused class into RGM in this case. I suggest either making it a runtime check (e.g, a flag), or removing the class from the codegen.

RobertBColton: Which class is the one you are saying is unused? Are you saying RGMPlugin is unused or ServerPlugin is unused? The former is clearly true, but the latter is not always true. I am in fact filtering out the ServerPlugin class entirely in both the pro qmake and the CMakeLists.

JoshDreamland: So if you remove the `#if` from just MainWindow, it will generate link errors because the Qt project file excludes the class?

RobertBColton: Yes, assuming you haven't uncommented `rgm_enable_grpc_server` in the pro or CMakeLists.

JoshDreamland: A bit awkward, but okay. Proceed.

RobertBColton: Silly me though I forgot to wrap the include in the main window.

RobertBColton: It's ok though, as I said later we'll move plugins to dlls.

PR #93, opened 2020-03-07
RGM Enable Syntax Highlight

RobertBColton:
Remember that fundies did not like how the CLI server option was called disabled and so he had it changed to enabled? Well, keeping in that tradition, I am inverting the RGM option for enabling scintilla. Now it's enable instead of disable (still disabled by default in the pro). This just means all of the options we provide across projects are semantically consistent.

### DIFF
--- a/RadialGM.pro
+++ b/RadialGM.pro
@@ -20,14 +20,14 @@ QMAKE_TARGET_PRODUCT = RadialGM IDE
 QMAKE_TARGET_DESCRIPTION = ENIGMA Development Environment
 QMAKE_TARGET_COPYRIGHT = "Copyright \\251 2007-2018 ENIGMA Dev Team"
 
-# Uncomment if you want QPlainTextEdit used in place of QScintilla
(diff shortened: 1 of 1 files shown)


RadialGM.pro
```diff
@@ -20,14 +20,14 @@ QMAKE_TARGET_PRODUCT = RadialGM IDE
 QMAKE_TARGET_DESCRIPTION = ENIGMA Development Environment
 QMAKE_TARGET_COPYRIGHT = "Copyright \\251 2007-2018 ENIGMA Dev Team"
 
-# Uncomment if you want QPlainTextEdit used in place of QScintilla
-CONFIG += rgm_disable_syntaxhighlight
+# Comment out if you want QPlainTextEdit used in place of QScintilla
```
JoshDreamland: The comment shouldn't invite people to take an action that's already be taken. I know it was like this before, but it's confusing.

RobertBColton: Alright: 08d1f0636feddaffabe67d0dc1329e79dfdd7941

PR #95, opened 2020-03-08
Azure Enable GRPC for Win32

RobertBColton:
Enable GRPC client for the Azure build on Win32.
https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html#module:CMakeDependentOption
Also fix the option itself by removing PARENT_SCOPE, which was actually setting the variable a scope above the current scope and thus causing it to not actually build the server plugin.
https://cmake.org/cmake/help/v3.3/command/set.html

I tested the artifact produced from this build and it does indeed have the GRPC client built in. However, when launched from MSYS2 it doesn't seem to launch emake like my local build does. Although, if emake is manually launched, the Azure artifact will still communicate with it.

That just means this pull request is correct and we just need to work out yet how we expect people to launch this thing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,8 +161,8 @@ endif()
 
 if (RGM_ENABLE_GRPC_SERVER)
   add_definitions(-DRGM_SERVER_ENABLED)
-  set(RGM_SOURCES ${RGM_SOURCES} Plugins/ServerPlugin.cpp PARENT_SCOPE)
-  set(RGM_HEADERS ${RGM_HEADERS} Plugins/ServerPlugin.h PARENT_SCOPE)
+  set(RGM_SOURCES ${RGM_SOURCES} Plugins/ServerPlugin.cpp)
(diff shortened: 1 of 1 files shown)


RobertBColton:
Actually, it's just allegedly broke on Linux, or specifically fundies machine. As I said above, I tested the artifact and it's working over here on Windows. Despite that, fundies never actually disabled GRPC from the CMake/Azure build, he disabled it from the pro file which he and I use locally. All I've basically done now is made an option to turn it on or off easily, with no difference from how fundies left it a few days ago.

RobertBColton:
Ok fundies wants a Windows check around this and to only turn it on for Windows right now. I had it backwards, I use the pro file, he uses CMake himself locally.

RobertBColton:
Alright, I added the check to only it for Win32 right now and it seems to work. I tested all 3 of the Windows releases and double checked that `ServerPlugin` was not getting built on Arch Linux like fundies wanted me to. With that said, I am merging this as a success!

fundies:
Its broken on josh's machine too las we checked so its not just a local issue for me

PR #101, opened 2020-03-12
actually disable grpc when no grpc & fixes grpc issues on linux

fundies:
partner of https://github.com/enigma-dev/enigma-dev/pull/1919


CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
 cmake_minimum_required(VERSION 3.11)
 
-include(CMakeDependentOption)
-cmake_dependent_option(RGM_ENABLE_GRPC_SERVER "Enable the GRPC client plugin for compilation and code analysis." ON "WIN32" OFF)
+if (WIN32)
+  option(RGM_ENABLE_GRPC_SERVER "Enable the GRPC client plugin for compilation and code analysis." ON)
+else()
+  option(RGM_ENABLE_GRPC_SERVER "Enable the GRPC client plugin for compilation and code analysis." OFF)
+endif()
```
RobertBColton: Why did you do this?

MainWindow.cpp
```diff
@@ -20,7 +20,10 @@
 #include "Components/Logger.h"
 
 #include "Plugins/RGMPlugin.h"
+
+#ifdef RGM_SERVER_ENABLED
 #include "Plugins/ServerPlugin.h"
+#endif
```
RobertBColton: Thanks, I did miss that one.

CMakeLists.txt
fundies: cause your way completely hides the option on linux mine just disables it

fundies: also, cause i dont have OCD like u

RobertBColton: Well it's not really OCD, it's just like if you have to type that string multiple times you are more likely to make a typo and cause a run time bug. Whereas if you make a typo in my way, even though it's more characters, it's instead more likely to be a compile time error.

fundies: sounds like ocd to me. if you tested that works ill append the pr tho

CMakeLists.txt
```diff
@@ -317,4 +317,4 @@ if (WIN32)
           fixup_bundle(\"${APPS}\" \"${LIBS}\" \"${SEARCH_PATHS}\")
           execute_process(COMMAND windeployqt.exe ${WINDEPLOY_ARGS} ${APPS})
           ")
-endif()
+endif()
```
RobertBColton: Missed end of line!

RobertBColton: I got it.

PR #105, opened 2020-03-15
yoink qtcreator's ansi handler

fundies:


MainWindow.cpp
```diff
@@ -139,7 +139,24 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainW
...
+        }
+        outputTextBrowser->textCursor().insertText(text.mid(startPos, crPos - startPos), format);
+        outputTextBrowser->textCursor().clearSelection();
+        outputTextBrowser->textCursor().movePosition(QTextCursor::StartOfBlock, QTextCursor::KeepAnchor);
+        startPos = crPos + 1;
+    }
+    if (startPos < text.count())
+        outputTextBrowser->textCursor().insertText(text.mid(startPos), format);
+  });
+
```
RobertBColton: If you hooked this to output text browser's insert signal or w/e then ANSI escapes would be colored for all plugins instead of just the server plugin.

Plugins/ServerPlugin.cpp
```diff
@@ -118,9 +119,13 @@ struct SystemReader : public AsyncReadWorker<SystemType> {
...
 struct CompileReader : public AsyncReadWorker<CompileReply> {
   virtual ~CompileReader() {}
+  Utils::AnsiEscapeCodeHandler ansiHandler;
   virtual void process(const CompileReply& reply) final {
     for (auto log : reply.message()) {
-      emit LogOutput(log.message().c_str());
+	  QList<Utils::FormattedText> txts = ansiHandler.parseText(QString::fromStdString(log.message() + "\n"));
+	  for (auto& str : txts) {
+	    emit LogOutput(str.text, str.format);
+	  }
```
RobertBColton: This ANSI escape stuff itself could have been made a plugin and more loosely coupled but I suppose that might be overkill.

Components/ANSIescapeCodeHandler.cpp
```diff
@@ -0,0 +1,248 @@
...
+*/
+
+AnsiEscapeCodeHandler::AnsiEscapeCodeHandler() :
+    m_previousFormatClosed(true)
+{
+}
+
+static QColor ansiColor(uint code)
+{
+    if (code < 8) return QColor();
```
RobertBColton: Use `R_EXPECT` here, albeit this tightly couples this component with the rest of our code base making it harder to pull updates to this source from Qt again.
https://github.com/enigma-dev/RadialGM/blob/b3cc0ac81a2011c300878b1f4893e92727e729f4/Components/Logger.h#L29

Components/ANSIescapeCodeHandler.cpp
```diff
@@ -0,0 +1,248 @@
...
+        outputData << FormattedText(input.text.left(escapePos), charFormat);
+    }
+
+    const QChar semicolon       = QLatin1Char(';');
+    const QChar colorTerminator = QLatin1Char('m');
+    const QChar eraseToEol      = QLatin1Char('K');
+    // strippedText always starts with "\e["
+    QString strippedText = input.text.mid(escapePos);
+    while (!strippedText.isEmpty()) {
+        while (strippedText.startsWith(escape)) {
```
RobertBColton: Looks like you dropped two other asserts from the original here.
https://github.com/qt-creator/qt-creator/blob/fc9bdf1255741f9261572d8560c6d8d1def76828/src/libs/utils/ansiescapecodehandler.cpp#L95
https://github.com/qt-creator/qt-creator/blob/fc9bdf1255741f9261572d8560c6d8d1def76828/src/libs/utils/ansiescapecodehandler.cpp#L104

Also, I like how `QTC_ASSERT` accepts break as an argument, we should do that maybe.

fundies: I think i just used older version

RobertBColton: What? Well that's probably why the color doesn't work!

Components/ANSIescapeCodeHandler.cpp
```diff
@@ -0,0 +1,248 @@
...
+            strippedText.remove(0, 2);
+
+            // \e[K is not supported. Just strip it.
+            if (strippedText.startsWith(eraseToEol)) {
+                strippedText.remove(0, 1);
```
RobertBColton: If you comment out just this single line you well see the letter k on both sides of text that you would expect to be colored. That means GCC is coloring its output using specific ANSI escapes which this ANSI escape handler does not support.

Plugins/ServerPlugin.cpp
RobertBColton: Actually I guess this is fine because now all plugins have ability to log formatted output of their own.

RobertBColton:
Ok after some testing, I don't have any idea how Qt Creator manages to keep the view at bottom without applying the text cursor like I do above. However, I dislike doing it this way because this is how LGM did it, and it doesn't let you scroll up at all while output is coming (very annoying!). I like the behavior of `QPlainTextEdit::append` where if you scroll up, the viewport stays there, but the text cursor continues moving down and appending new text.
https://code.qt.io/cgit/qt/qtbase.git/tree/src/widgets/widgets/qplaintextedit.cpp?h=dev#n3047

This is much closer to how a terminal actually behaves and in fact is exactly what my testing shows in MSYS2 for me.

JoshDreamland:
If fundies insists, Robert, let him use this. I think it's ugly as stop-gaps go, but it'll work.

Another alternative is this library: https://github.com/theZiz/aha/blob/master/aha.c

Seems we could copy-paste that instead. It's MPL-licensed, which is closer to LGPL. It might support that weird `\e[k` thing. And it feels like it's an easier library to contribute back to if it has shortcomings.

RobertBColton:
That weird `\e[K` thing is exactly the reason the color doesn't actually work, only bold does.

Issue #96, opened 2020-03-08
High DPI Problems


RobertBColton:
There's going to be a load of these but this is the first bothering me on my 4k display. The tree view has inconsistent icon sizes. The disabled pixmap on empty groups is being rendered bigger than the icons used on all the other tree items. Part of the problem (though I don't understand why it is a problem) is just that we use the disabled pixmap from the group icon.
https://github.com/enigma-dev/RadialGM/blob/8960e221a81ef2cafae39461cd3fa4b6722a6b52/Models/TreeModel.cpp#L85
This may not actually be our fault because I have tested using a QIcon vs a QPixmap in a tree model, and I found that icons are always rendered smaller than pixmaps even if they both are loaded from the exact same file. For that reason I am considering reporting this bug upstream to Qt. One workaround for this issue I found is to just construct and return an icon from the disabled pixmap, or alternatively we could change all the icons to pixmaps and use only pixmaps.

| QIcon(":/resources/group.png" ) | QPixmap(":/resources/group.png" ) |
|---------------------------------|-----------------------------------|
|<img src="https://user-images.githubusercontent.com/3212801/76171075-ee888a80-615d-11ea-9376-c3bb2462b051.png" width="250">|<img src="https://user-images.githubusercontent.com/3212801/76171068-e03a6e80-615d-11ea-987e-87fd9f51a6f1.png" width="250">|

RobertBColton:
I made a test project that reproduces it so I can file it upstream. It occurs with high dpi simply being enabled.
```cpp
QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
```

![High DPI QPixmap Bigger Than QIcon](https://user-images.githubusercontent.com/3212801/76183611-6925ca00-619f-11ea-800f-8914ad535443.png)

RobertBColton:
Filed upstream: https://bugreports.qt.io/browse/QTBUG-82738

PR #115, opened 2020-04-05
Tree Folder DPI Fix

RobertBColton:
This is my attempt to work around #96 since Qt closed my bug report. I wrap the disabled folder pixmap in a QIcon, which does scaling, so that this model returns nothing but QIcons. The alternative would be to change all of the uses of QIcon to QPixmap, and return only pixmaps, but pixmaps are not scaled automatically. The other alternative is we write our own tree view delegate that actually fixes the Qt bug report I filed.

To me this seems like a fine work around, I'm just bothered by the tree having icons smaller than toolbars and other UI elements.

| Before                                                                                                                         | After                                                                                                                          |
|--------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
| Tree Varying Icon Size | Tree Uniform Icon Size |

### DIFF
--- a/Models/TreeModel.cpp
+++ b/Models/TreeModel.cpp
@@ -82,7 +82,7 @@ QVariant TreeModel::data(const QModelIndex &index, int role) const {
 
     const QIcon &icon = it->second;
     if (item->type_case() == TypeCase::kFolder && item->child_size() <= 0) {
-      return icon.pixmap(icon.availableSizes().first(), QIcon::Disabled);
+      return QIcon(icon.pixmap(icon.availableSizes().first(), QIcon::Disabled));
     }
     return icon;
   } else if (role == Qt::EditRole || role == Qt::DisplayRole) {


RobertBColton:
No, it isn't Josh, see long and detailed discussion me and Rusky had. What happens is the styled item delegate through the tree view is asking the tree model for a pixmap and then scaling it.

JoshDreamland:
And before it was... not scaling it? Just showing a tiny thing?

You're still asking for the smallest available size of the icon.

RobertBColton:
Right now, there's only one size of the icon, 18px * 18px, until we do SVG icons or something or another. For all intents and purposes, yes that's the smallest size, but it's also the largest.

Regardless, you were half correct, it is later wrapped in a QIcon further up.
https://code.woboq.org/qt5/qtbase/src/widgets/itemviews/qstyleditemdelegate.cpp.html#336

The problem is that QStyledItemDelegate uses different scaling for QPixmap than it does QIcon. That's how you get the varying sized icons I showed in #96 which we want fixed.

PR #116, opened 2020-04-10
Room View Spatial Hash

### LINKED ISSUE #122: Default Values
This is really odd and occurs on master. For some reason the x and y scale are invalid when loading a GMX, but the x and y value of instances is correct. I can reproduce it loading TKG's Key to Success GMX project. I looked high and low and can't seem to find the source of the bug, the Room proto seems correct and has the correct GMX alias of the field name.

RobertBColton:
This adds spatial hashing to the room editor for more efficient repainting.

Left to do:
- [ ] Add a hash for tiles.
- [ ] Use full instance/tile transform in proxy.
- [ ] Handle model data changes, e.g, depth/x/y/bounding box.
- [ ] Merge sort the query window results by depth.
- [x] Resolve #122 which is related to the proxy.

### DIFF
--- a/Widgets/AssetScrollAreaBackground.cpp
+++ b/Widgets/AssetScrollAreaBackground.cpp
@@ -184,11 +184,11 @@ void AssetScrollAreaBackground::paintEvent(QPaintEvent* /* event */) {
     _totalDrawOffset = GetCenterOffset() + _userDrawOffset;
 
     painter.save();
-    QRect visible(-this->pos(), ((QWidget*)this->parent())->size());
+    QRect clip(-this->pos(), ((QWidget*)this->parent())->size());
+    painter.setClipRect(clip);
     painter.translate(_totalDrawOffset);
     painter.scale(_currentZoom, _currentZoom);
-    visible = painter.transform().inverted().mapRect(visible);
-    _assetView->Paint(painter, visible);
+    _assetView->Paint(painter);
     painter.restore();
 
     GridDimensions g = _assetView->GetGrid();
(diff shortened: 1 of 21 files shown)


Widgets/AssetScrollAreaBackground.cpp
```diff
@@ -184,6 +184,8 @@ void AssetScrollAreaBackground::paintEvent(QPaintEvent* /* event */) {
     _totalDrawOffset = GetCenterOffset() + _userDrawOffset;
 
     painter.save();
+    QRect clip(-this->pos(), ((QWidget*)this->parent())->size());
+    painter.setClipRect(clip);
```
RobertBColton, line 188: This was made possible by #120 which sets the clip on master.

Widgets/RoomView.h
```diff
@@ -14,6 +17,57 @@ class InstanceSortFilterProxyModel : public QSortFilterProxyModel {
...
+};
+
+struct InstanceProxy : Proxy {
+  InstanceSortFilterProxyModel *model;
+  InstanceProxy(InstanceSortFilterProxyModel *model, int row): Proxy(row), model(model) {}
+  int x1() const override { return model->data(model->index(row, Room::Instance::kXFieldNumber)).toInt(); }
+  int y1() const override { return model->data(model->index(row, Room::Instance::kYFieldNumber)).toInt(); }
+  int x2() const override {
+    int w = sprw();
+    double xscale = model->data(model->index(row, Room::Instance::kXscaleFieldNumber)).toDouble();
```
RobertBColton, line 44: This is effected by #122 which is causing the spatial hash to be glitchy unless manually patched over.

PR #120, opened 2020-05-04
Asset View Clip

### LINKED ISSUE #118: Asset Background Problems
I need to document a couple of issues that came up during implementation of the spatial hash for the room editor. They pertain mainly to the clipping of the painting for asset scroll area.

1) We are painting the room grid outside of the actual room, which LGM does not do, but I am totally fine with this. I think it's a good idea.
2) AssetScrollAreaBackground is not using its paint event at all, which tells it the current region being exposed that needs to be repainted. It's much more efficient for us to only paint what needs to be painted.
3) AssetView is not informed of the paint event by AssetScrollAreaBackground and thus has to completely guess which part of it is exposed for painting. This makes it difficult for the room view to know which window to query in the spatial hash.

RobertBColton:
Alright, now I've figured out how to set the clip to just the visible viewport of the scroll area, like we wanted. I use the asset view scroll area background's negated position to find the scroll offset of the visible viewport. I use the size of the asset view scroll area background's parent, which only includes scrollbars when they are visible, as the size of the clipping rectangle. I then apply this clip to the painter before it is passed to the asset view, in this case the room view.

I see no regressions with TKG's game Key to Success or with other games. Zooming appears to work correctly and I also double checked the clip by filling it as a red rectangle. I still recommend that maybe we try this out on a few other games first. Otherwise, we should be safe to use this clip to resolve #118 so that the room view knows the appropriate window of its spatial hash to query.

### DIFF
--- a/Widgets/AssetScrollAreaBackground.cpp
+++ b/Widgets/AssetScrollAreaBackground.cpp
@@ -184,6 +184,8 @@ void AssetScrollAreaBackground::paintEvent(QPaintEvent* /* event */) {
     _totalDrawOffset = GetCenterOffset() + _userDrawOffset;
 
     painter.save();
+    QRect clip(-this->pos(), ((QWidget*)this->parent())->size());
+    painter.setClipRect(clip);
     painter.translate(_totalDrawOffset);
     painter.scale(_currentZoom, _currentZoom);
     _assetView->Paint(painter);


RobertBColton:
This corrects the overpainting that we are doing. There's no sense in painting what isn't visible.

RobertBColton:
After some extensive benchmarking and researching by Rusky and I, we have decided there's no merit in this clip. We timed our existing room painting code with and without the clip, which yielded no significant differences. After the discovery of #132 we also tested this out with a static pixmap, randomly generated positions, and with and without a clip and found no differences. We even theorized a bit about small vs large clip sizes, assuming at some point Qt's internal clipping of the newly exposed area was doing a better job than we were. We also looked at try the painter window and view, which only made things worse. So I am going to close this and as far as the spatial hash in #116 is concerned, we can simply pass the query window as an extra parameter.

PR #123, opened 2020-05-06
Default Model Values

### LINKED ISSUE #122: Default Values

RobertBColton:
This is how Josh would like to resolve #122 by making it explicit when you are ok accepting the default value from the model. I had to also allow column 0 as a special sentinel for `RepeatedMessageModel` to return a pointer to the requested model so that I have an actual way to call `dataOrDefault` on repeated messages. The hack avoids the complication of having to add `dataOrDefault` on all the higher level models and interfaces, such as the sort filter proxy model.

### DIFF
--- a/Models/MessageModel.cpp
+++ b/Models/MessageModel.cpp
@@ -91,7 +91,7 @@ QVariant MessageModel::Data(int row, int column) const {
   return data(this->index(row, column, QModelIndex()), Qt::DisplayRole);
 }
 
-template<bool HasField>
+template<bool NO_DEFAULT>
 QVariant MessageModel::dataInternal(const QModelIndex &index, int role) const {
   R_EXPECT(index.isValid(), QVariant()) << "Supplied index was invalid:" << index;
   if (role != Qt::DisplayRole && role != Qt::EditRole && role != Qt::DecorationRole) return QVariant();
@@ -132,7 +132,7 @@ QVariant MessageModel::dataInternal(const QModelIndex &index, int role) const {
   }
 
   // If the field has't been initialized return an invalid QVariant. (see QVariant.isValid())
-  if (HasField && !refl->HasField(*_protobuf, field)) return QVariant();
+  if (NO_DEFAULT && !refl->HasField(*_protobuf, field)) return QVariant();
 
   switch (field->cpp_type()) {
     case CppType::CPPTYPE_MESSAGE: R_EXPECT(false, QVariant()) << "The requested field " << index << " is a message";
--- a/Models/MessageModel.cpp
+++ b/Models/MessageModel.cpp
@@ -91,23 +91,8 @@ QVariant MessageModel::Data(int row, int column) const {
   return data(this->index(row, column, QModelIndex()), Qt::DisplayRole);
 }
 
-QVariant MessageModel::data(const QModelIndex &index, int role) const {
-  QVariant ret = dataOrDefault(index, role);
-  if (!ret.isValid()) return QVariant();
-
-  const Descriptor *desc = _protobuf->GetDescriptor();
-  const Reflection *refl = _protobuf->GetReflection();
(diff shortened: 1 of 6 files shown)


Models/MessageModel.cpp
```diff
@@ -92,6 +92,22 @@ QVariant MessageModel::Data(int row, int column) const {
...
 
 QVariant MessageModel::data(const QModelIndex &index, int role) const {
+  QVariant ret = dataOrDefault(index, role);
+  if (!ret.isValid()) return QVariant();
+
+  const Descriptor *desc = _protobuf->GetDescriptor();
+  const Reflection *refl = _protobuf->GetReflection();
+  const FieldDescriptor *field = desc->FindFieldByNumber(index.row());
+
+  if (!field) return ret; // TODO: Wipe out table hack below in dataOrDefault.
```
RobertBColton, line 102: Could possibly fix this using custom role.

Models/MessageModel.cpp
```diff
@@ -131,7 +132,7 @@ QVariant MessageModel::data(const QModelIndex &index, int role) const {
   }
 
   // If the field has't been initialized return an invalid QVariant. (see QVariant.isValid())
-  if (!refl->HasField(*_protobuf, field)) return QVariant();
+  if (HasField && !refl->HasField(*_protobuf, field)) return QVariant();
```
RobertBColton, line 135: Rusky says we should treat this as an error/with an assert.

Models/RepeatedMessageModel.cpp
```diff
@@ -41,5 +41,7 @@ bool RepeatedMessageModel::setData(const QModelIndex &index, const QVariant &val
 }
 
 QVariant RepeatedMessageModel::data(const QModelIndex &index, int role) const {
+  // protobuf field number with 0 is impossible, use as sentinel to get model itself
+  if (index.column() == 0) return QVariant::fromValue(_subModels[index.row()]);
```
RobertBColton, line 45: I don't think there's any implications of this?

Widgets/RoomView.cpp
```diff
@@ -85,18 +85,16 @@ void RoomView::paintTiles(QPainter& painter) {
     bkg = bkg->GetSubModel<MessageModel*>(TreeNode::kBackgroundFieldNumber);
     if (!bkg) continue;
 
+    MessageModel* tile = _sortedTiles->data(_sortedTiles->index(row, 0)).value<MessageModel*>();
```
RobertBColton, line 88: Should we log/check this? Also, should we check if the QVariant is valid before taking the value?

Models/RepeatedMessageModel.cpp
```diff
@@ -37,9 +44,18 @@ int RepeatedMessageModel::columnCount(const QModelIndex & /*parent*/) const {
...
+    "Supplied row was out of bounds:" << index.row();
   return _subModels[index.row()]->setData(_subModels[index.row()]->index(index.column()), value, role);
 }
 
 QVariant RepeatedMessageModel::data(const QModelIndex &index, int role) const {
+  R_EXPECT(index.row() >= 0 && index.row() < _subModels.size(), QVariant()) <<
+    "Supplied row was out of bounds:" << index.row();
+  // protobuf field number with 0 is impossible, use as sentinel to get model itself
+  if (index.column() == 0) {
+    R_EXPECT(index.isValid(), QVariant()) << "Supplied index was invalid:" << index;
```
RobertBColton, line 57: This one might be redundant because it just checks that the row and column are non-negative. We already check the row, because it's necessary to do so. And the column would not be negative if it was equal to 0, so I don't really think this is necessary.

PR #128, opened 2020-06-18
Update VCPKG

RobertBColton:
This is supposed to include a fix for the MSVC build's icu vcpkg issue.


fundies:
Looks like google's hack fuxed build too. Can you try commenting this line: https://github.com/enigma-dev/RadialGM/blob/master/CMakeLists.txt#L11 and see if that helps any?

RobertBColton:
I recreated it as #129 so that it has the DLLEXPORT fixes we just merged.

RobertBColton:
It's also not going to work until #130 is merged.

RobertBColton:
Alright with #130 merged, I created #131 that has the Arch fix.

Issue #132, opened 2020-06-19
Data Query is Slow


RobertBColton:
Discovered this while benchmarking the room clip with Rusky. Basically, accessing proto data through the models we provide is pretty damn slow, even in release mode. While this may be ok for some editors and reflection based views, it's particularly problematic for performance-critical views like the room editor.

One area I tested was caching of the descriptor and reflector used in the proto model. This cut field access for the x property of 100,000 instances from 13,000~ microseconds down to about 6,000~ microseconds.

https://github.com/enigma-dev/RadialGM/blob/faff910d6398ab2660f45c512450f48d45b47099/Models/MessageModel.cpp#L99

I have not yet been able to determine from the documentation whether it would be ok for us to do this all the time.
https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Reflection

There are also other potential avenues for optimizing all of this, including avoidance of QVariant for data access. One way to do so would be to simply grab and use the protobuf directly in the room painting, rather than access the data through the model.

fundies:
13000 micro seconds is fast enough :/. The room editor only updates on changes we're not trying to draw it 60x/sec

JoshDreamland:
It's fast enough... as long as we're caching this data somewhere else for things like animations. And we're going to want animations.

It's telling if we're saturating a core rendering something at normal game speeds. I think the problem is the reflection; even proto caches data, normally.

To fundies' point, the data won't update all that often. We can probably get away with a total-complete destructive re-cache on update. But I don't think we have to. I think if you just key the fields you want by the rows they came from, you'll be fine.

This logic doesn't even have to be specific to the room editor, but for now, it probably should be. There's probably higher-priority work, but I wouldn't tell someone not to tackle a problem they're interested in.

RobertBColton:
You guys are missing a very important thing about what I said, I isolated only the x property, but there are about 10 properties plus a sprite lookup by string to draw every instance. So in reality, just to query the data necessary to draw 100,000 instances, not to even draw them, probably takes about 13,000 * 10 or 130,000 microseconds. And you're still both ignoring that caching of the descriptor and the reflector in the proto model cut it in half, which is a very good optimization!

JoshDreamland:
Everything we were discussing stands after multiplying the time by five and dividing by two.

A 2x speedup is good, but it won't cut it. We'll need more caching if we mean to render animations.

fundies:
we could make it faster by removing all functionality

PR #133, opened 2020-06-23
Proto Model Reflection Cache

RobertBColton:
This is the first step in optimizing #132 by caching the reflector and descriptors used by the proto based models. I think we will need to do additional testing and benchmarking to ensure that we can do this safely.

I have found from the documentation already that seems it should be appropriate to cache the reflection object so long as we don't violate any of the specific rules it mentions. My analysis is that we certainly don't, because we are obviously never using a reflection object on a message whose type is not compatible.
https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Reflection

### DIFF
--- a/Models/MessageModel.cpp
+++ b/Models/MessageModel.cpp
@@ -11,10 +11,8 @@ MessageModel::MessageModel(ProtoModel *parent, Message *protobuf) : ProtoModel(p
 MessageModel::MessageModel(QObject *parent, Message *protobuf) : ProtoModel(parent, protobuf) { RebuildSubModels(); }
 
 void MessageModel::RebuildSubModels() {
-  const Descriptor *desc = _protobuf->GetDescriptor();
-  const Reflection *refl = _protobuf->GetReflection();
-  for (int i = 0; i < desc->field_count(); i++) {
-    const FieldDescriptor *field = desc->field(i);
+  for (int i = 0; i < _desc->field_count(); i++) {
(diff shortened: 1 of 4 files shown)


RobertBColton:
I did a little test that seems to confirm that descriptors and reflectors for messages of the same type are in fact the same.

```cpp
  qDebug() << protobuf->GetTypeName().c_str()
           << protobuf->GetDescriptor()
           << protobuf->GetReflection();
```
```
Debug in ../RadialGM/Models/ProtoModel.cpp:11 aka ProtoModel::ProtoModel(QObject*, google::protobuf::Message*):
	buffers.TreeNode 0xa6f63f0 0x130f3b10
Debug in ../RadialGM/Models/ProtoModel.cpp:11 aka ProtoModel::ProtoModel(QObject*, google::protobuf::Message*):
	buffers.TreeNode 0xa6f63f0 0x130f3b10
Debug in ../RadialGM/Models/ProtoModel.cpp:11 aka ProtoModel::ProtoModel(QObject*, google::protobuf::Message*):
	buffers.TreeNode 0xa6f63f0 0x130f3b10
Debug in ../RadialGM/Models/ProtoModel.cpp:11 aka ProtoModel::ProtoModel(QObject*, google::protobuf::Message*):
	buffers.TreeNode 0xa6f63f0 0x130f3b10
```

RobertBColton:
This is dead because #144 brings the super model, it can be cached there later.

PR #134, opened 2020-06-25
MSYS2 Dash Azure Workaround

RobertBColton:
This is a small workaround to fix the broken MinGW32 and MinGW64 builds on our Azure CI right now.

I tried the safer fix mentioned, but it didn't work, so I had to go with the filesystem hack.
https://github.com/msys2/MSYS2-packages/issues/2021

It's probably not worth merging this because anybody who does a fresh install of MSYS2 and then ENIGMA might have this problem anyway? @fundies ?


fundies:
Right, This is just another mingw bug caused by alex. Hopefully it will remedy itself in week or so. So I would just leave it alone for now

RobertBColton:
Closing because it was already automatically fixed.

PR #137, opened 2020-07-23
Keybinding Preferences

RobertBColton:
image


Dialogs/PreferencesDialog.cpp
```diff
@@ -42,6 +42,19 @@ void PreferencesDialog::setupKeybindingUI() {
 
     ui->keybindingTree->addTopLevelItem(item);
   }
+
+  connect(ui->buttonBox->button(QDialogButtonBox::Reset), &QAbstractButton::clicked, this,
```
fundies, line 46: why you using connect instead of the auto slot or w/e ```onWhateverClicked() {}```

PR #138, opened 2020-07-25
egm supp

fundies:


MainWindow.cpp
```diff
@@ -73,15 +77,44 @@ void diagnosticHandler(QtMsgType type, const QMessageLogContext &context, const
...
+  } else {
+    qDebug() << "Error: Failed to locate ENIGMA sources. Loading internal events.ey.\n" << "Search Paths:\n" << MainWindow::EnigmaSearchPaths;
+    QFile internal_events(":/events.ey");
+    internal_events.open(QIODevice::ReadOnly | QFile::Text);
+    std::stringstream ss;
+    ss << internal_events.readAll().toStdString();
+    _event_data = std::make_unique<EventData>(ParseEventFile(ss));
+  }
+  
+  egm = egm::EGM(_event_data.get());
```
RobertBColton, line 116: A lot of this should probably go into its own project handling class, like Josh had me do for recent files. Jamming everything into the main window class is not a good, we want to design this with loose coupling of components, something that signals and slots help accomplish.

MainWindow.cpp
```diff
@@ -155,7 +188,7 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainW
   openNewProject();
 }
 
-MainWindow::~MainWindow() { delete _ui; }
+MainWindow::~MainWindow() { diagnosticTextEdit = nullptr; delete _ui; }
```
RobertBColton, line 191: This isn't necessary it uses the Qt ownership mechanism (it's parent is specified in the constructor). This is C++, not C, we're not concerned about zero'ing out our pointers.

https://doc.qt.io/qt-5/objecttrees.html

MainWindow.cpp
```diff
@@ -302,7 +337,7 @@ void MainWindow::openFile(QString fName) {
 
   MainWindow::setWindowTitle(fileInfo.fileName() + " - ENIGMA");
   _recentFiles->prependFile(fName);
-  openProject(std::unique_ptr<buffers::Project>(loadedProject));
+  openProject(std::move(loadedProject));
```
RobertBColton, line 340: Should document somewhere, maybe in the header, that `openProject` takes ownership of the buffer.

MainWindow.h
```diff
@@ -108,11 +114,15 @@ class MainWindow : public QMainWindow {
...
   QPointer<RecentFiles> _recentFiles;
+  
+  std::unique_ptr<EventData> _event_data;
+  egm::EGM egm;
 
   void openSubWindow(buffers::TreeNode *item);
   void readSettings();
   void writeSettings();
   void setTabbedMode(bool enabled);
+  static QFileInfo getEnigmaRoot();
```
RobertBColton, line 125: This is a horrible function name, also add comments to clarify what it does.

MainWindow.h
```diff
@@ -108,11 +114,15 @@ class MainWindow : public QMainWindow {
 
   std::unique_ptr<buffers::Project> _project;
   QPointer<RecentFiles> _recentFiles;
+  
+  std::unique_ptr<EventData> _event_data;
+  egm::EGM egm;
```
RobertBColton, line 119: Move all of this stuff to the private section.

images.qrc
```diff
@@ -100,5 +100,6 @@
         <file alias="banner.png">Images/banner.png</file>
         <file alias="icon.ico">Images/icon.ico</file>
         <file alias="transparent.png">Images/transparent.png</file>
+        <file alias="events.ey">Submodules/enigma-dev/events.ey</file>
```
RobertBColton, line 103: Maybe we should rename images.qrc or resources.qrc because the user will later replace icon packs dynamically. We don't want them to have to put `events.ey` into their icon pack.

MainWindow.cpp
```diff
@@ -73,15 +77,44 @@ void diagnosticHandler(QtMsgType type, const QMessageLogContext &context, const
...
+    }
+  }
+  
+  return EnigmaRoot;
+}
+
+MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), _ui(new Ui::MainWindow), _event_data(nullptr), egm(nullptr) {
+
+  if (!EnigmaRoot.filePath().isEmpty()) {
+    _event_data = std::make_unique<EventData>(ParseEventFile((EnigmaRoot.absolutePath() + "/events.ey").toStdString()));
```
RobertBColton, line 106: Ask @JoshDreamland why `ParseEventFile` doesn't just return a `unique_ptr` for me.

MainWindow.cpp
fundies: The editor segfaults on close without this. You're referencing another pointer so it's never cleared.

MainWindow.h
fundies: It gets the root of enigma :/ I dunno what else you'd name it.

MainWindow.h
fundies: It is in private....

images.qrc
fundies: IRight, I was going to rename this file but didn't care to do it here

MainWindow.cpp
fundies: EventData should be an object not a pointer but because it doesn't have a default constructor I need to conditionally initialize it with different constructors by making it a pointer.

MainWindow.cpp
fundies: Sounds like a task for another day.

MainWindow.cpp
RobertBColton: I think it's the signals queueing messages, something else is wrong, this is only a bandaid. The correct fix is to disconnect the output/log signals so it stops trying to reference that after it's gone.

MainWindow.h
RobertBColton: ENIGMA root could be anything, is my point, it could be interpreted as meaning the protobuf/tree root or etc. Probably `GetEnigmaFileSystemRoot` would be miles better.

RadialGM.pro
```diff
@@ -178,4 +178,5 @@ FORMS += \
 RESOURCES += \
     images.qrc
 
-DISTFILES +=
+DISTFILES += \
+  Submodules/enigma-dev/events.ey
```
RobertBColton, line 182: I don't know if this was necessary, the documentation says it only works on Unix, and you've already put it in the resources file.
https://stackoverflow.com/a/38102991

images.qrc
RobertBColton: I really wouldn't put it in with the images.
https://doc.qt.io/qt-5/resources.html#external-binary-resources

MainWindow.cpp
RobertBColton: This is new code, all _new_ code should be written correctly.

MainWindow.cpp
fundies: it also correct to null the pointer

MainWindow.h
fundies: none of the things you listed are enigma

RadialGM.pro
fundies: qtcreator did this not me

MainWindow.cpp
fundies: This is just the first draft to make it work/build against enigma. As long as you hold up this pr over nonsense design choices the CI remains broken. This pr was meant to be merged the same time as the enigma side one. You can refactor it later but it needs to go in now.

RadialGM.pro
RobertBColton: Hrm idk then.

Issue #20, opened 2018-03-31
Protocol Model


RobertBColton:
It's important that we get this worked out sooner rather than later, and I'd prefer the work to be done starting at the head of the grpc branch.

# How things work now
Every editor is constructed with a proto model and a single mapper that can only read and write from the resource associated with that editor:
https://github.com/enigma-dev/RadialGM/blob/9d1f6bcbcbde51739d1aaa0b7e49c4e81bff3ba5/MainWindow.cpp#L55
This allows editor construction to be generic and a map of the model for each resource (with its tree node as the key) to be held weakly by the main window.

# The problem with the current approach
The main window currently has a crash from the destruction of the QHash happening before the `QObject::destroyed` signal is received by the mdi subwindows:
https://github.com/enigma-dev/RadialGM/blob/9d1f6bcbcbde51739d1aaa0b7e49c4e81bff3ba5/MainWindow.cpp#L97

Some people feel this is a mistake with [the Qt documentation which explains](http://doc.qt.io/qt-5/qobject.html#destroyed) as though the `::destroyed` signal is processed before the destructor, when it's actually the opposite. So the QHash is gone when we get the signal and we can't remove it from the map then.

Not only is this a lot of pointer madness, it doesn't even allow a potential Room editor to access Object or Sprite data. Clearly we can just overload the constructor to the room editor, but then there would go our generic editor construction.

# What we should do instead
Let's combine all of the individual proto models into a table-like model and hold a unique_ptr to one instance of it in the main window. Not only is this a lot less nonsensical, and safer, it allows editors to refer to other resources without having to overload their constructors.

Specifically, the editor factory should be changed to:
```cpp
T *EditorFactory(ProtoModel *model, QModelIndex &modelIndex, QWidget *parent) { 
```
All editor constructors should be changed to:
```cpp
BaseEditor::BaseEditor(ProtoModel* model, QModelIndex &modelIndex, QWidget* parent)
: QWidget(parent), model(model), modelIndex(modelIndex), mapper(new ImmediateDataWidgetMapper(this)) {
```
The `ProtoModel` class should have a few public helper methods with the following prototypes:
```cpp
QModelIndex indexOfName(QString resourceName);
QModelIndex indexOfId(buffers::resources::TreeNode::TypeCase type, int id);
```

[...]

PR #144, opened 2020-07-30
Super Model

### LINKED ISSUE #139: TreeModel needs refactor
The tree model should refactored to inherit from the other models (particularly RepeadedMessageModel)

RobertBColton:
This is a complete and total rewrite to do what #20 had proposed and resolve #139 and its ilk forever. Rather than having a bunch of repeated duplicate models with broken logic scattered all over the place, we reintroduce ProtoModel as a giant recursive super model from which the entire project can be introspected and edited. The proto model becomes one [single source of truth](https://en.wikipedia.org/wiki/Single_source_of_truth) to be reckoned with.

* Drag and drop becomes "free" for all of the editors, as the super model handles it generically for any fields.
* Only a single implementation of the main model mutation logic.
* Updates to the project are reflected across the entire editor.
* The entire project can be introspected and edited from a single diagnostic view.
* Multi-project workspaces become a lot easier to do later on.
* This makes other features like popout editor windows easier too.

Diagnostic Inspector
Inspect Sound
Diagnostic Help Menu


JoshDreamland:
Let's revisit this. I need to sit down and piece together what's going on here. I'm told it's unfinished, but I like the 1KLoC reduction in size, for sure. Resurrecting this may be more efficient than what I was about to do (might end up going down the same rabbit holes).

RobertBColton:
The main issue here Josh is with change propagation. While the current/other way has issues propagating changes up and out to the rest of the editor, this way has the problem that it propagates changes too often. Something as simple as changing the name of a folder, would recursively signal everything below it as changed. That was why I decided that the current approach of model composition was best, because it's easier to deal with bubbling up than bubbling down.

In other words, this pull request is essentially proposing Vuex for Qt. They talk about this in their intro video.
https://vuex.vuejs.org/

PR #148, opened 2020-07-31
Drag and Drop Path Sort

RobertBColton:
This is another small tweak I walked across in my research of the super model. Our sort in `mimeData()` acts as though all of the indexes we are dragging are of the same parent which causes them to become interspersed when dropped. I realized I could solve that with a custom comparator which this pull request does. I'm also still not sure why we sort during drag instead of drop, although I recall that being the way it was in the official Qt example. Master currently behaves like LateralGM and GM does not have multiple selection for the tree.

Anyway, create some paths and scripts in the subfolders and select them, then drag them to some other folder and drop. If you do this on master you will get the master behavior below. If you do this on this pull request you will get the pull behavior below.
**NOTE**: If you double click the drop folder, it will change the selection, clicking the expand arrow does not!
| Select | Master | Pull |
|--------|--------|------|
|Select Paths & Scripts|Drop Paths & Scripts Master|Drop Paths & Scripts Pull|

### DIFF
--- a/Models/TreeModel.cpp
+++ b/Models/TreeModel.cpp
@@ -174,7 +174,14 @@ QMimeData *TreeModel::mimeData(const QModelIndexList &indexes) const {
 
   // rows are moved starting with the lowest so we can create
   // unique names in the order of insertion
-  std::sort(nodes.begin(), nodes.end(), std::less<QModelIndex>());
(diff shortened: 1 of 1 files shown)


RobertBColton:
It looks like this was fixed by the new tree model which sorts in mimeData.

PR #153, opened 2020-08-17
Object Editor Events

fundies:
image


MainWindow.h
```diff
@@ -114,8 +115,8 @@ class MainWindow : public QMainWindow {
 
   std::unique_ptr<buffers::Project> _project;
   QPointer<RecentFiles> _recentFiles;
-  
-  std::unique_ptr<EventData> _event_data;
+
+  static std::unique_ptr<EventData> _event_data;
```
RobertBColton, line 119: There's already a `static MainWindow::_instance` with a getter somewhere that I added. You can use it to refer to the main window and stuff for now. Please use it as it makes it easier to refactor the static access of the main window later on if we ever need to.

Models/EventsListModel.cpp
```diff
@@ -0,0 +1,82 @@
...
+QModelIndex EventsListModel::parent(const QModelIndex& index) const {
+  return sourceModel()->parent(index);
+}
+
+int EventsListModel::rowCount(const QModelIndex &parent) const {
+  return sourceModel()->rowCount(parent);
+}
+
+int EventsListModel::columnCount(const QModelIndex &parent) const {
+  return (parent.isValid()) ? 0 : 1;
```
RobertBColton, line 73: This is wrong where you keep doing it, stop doing it everywhere, the column count is always just 1 for _list_ models. Only in regards to the parent though, not in cases where you do actually want a variable number of columns.

>The columnCount() function is implemented for interoperability with all kinds of views, but by default informs views that the model contains only **one** column.
https://doc.qt.io/qt-5/qabstractlistmodel.html

>In most subclasses, the number of columns is **independent** of the parent.
https://doc.qt.io/qt-5/qabstractitemmodel.html#columnCount

```suggestion
return 1;
```

Models/EventsListModel.cpp
```diff
@@ -0,0 +1,82 @@
...
+QModelIndex EventsListModel::index(int row, int column, const QModelIndex& /*parent*/) const {
+  return createIndex(row, column);
+}
+
+QModelIndex EventsListModel::parent(const QModelIndex& index) const {
+  return sourceModel()->parent(index);
+}
+
+int EventsListModel::rowCount(const QModelIndex &parent) const {
+  return sourceModel()->rowCount(parent);
```
RobertBColton, line 69: This is wrong too, you can't pass proxy indexes to the source model, that's what the map functions exist for. The default implementation of proxy model does this automatically for you, so since you don't need to do anything special you can just delete this and similar overrides.

```suggestion
int EventsListModel::rowCount(const QModelIndex &parent) const {
  return sourceModel()->rowCount(mapToSource(parent));
}
```

RobertBColton: Since the base `QAbstractProxyModel` doesn't provide a default `::rowCount` I think you should just inherit `QIdentityProxyModel` instead since you're duplicate like 90% of what it does.

Dialogs/EventArgumentsDialog.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+const QStringList &EventArgumentsDialog::GetArguments() const { return arguments_; }
+
+void EventArgumentsDialog::done(int r) {
+  for (const QWidget *w : widgets_) {
+    QVariant argument = w->metaObject()->userProperty().read(w);
+    QString argstr = "";
+    if (QString(w->metaObject()->className()) == "QSpinBox") {
+      argstr = QString::number(reinterpret_cast<const QSpinBox *>(w)->value());
+    } else {
```
RobertBColton, line 101: Why did you do that? It should have worked the way it was, the user property of a spinbox is the value.
`Q_PROPERTY(int value READ value WRITE setValue NOTIFY valueChanged USER true)`
https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qspinbox.h.html#62

Dialogs/EventArgumentsDialog.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include <QDialogButtonBox>
+#include <QGridLayout>
+#include <QLabel>
+#include <QLineEdit>
+#include <QMetaProperty>
+#include <QPushButton>
+#include <QSpinBox>
+#include <QToolButton>
+
+EventArgumentsDialog::EventArgumentsDialog(QWidget *parent, const QStringList &arguments) : QDialog(parent) {
```
RobertBColton, line 16: Why are the arguments string and not QVariant or something. Why aren't we using real reflection over the event proto here.

Models/EventTypesListSortFilterProxyModel.cpp
```diff
@@ -0,0 +1,24 @@
...
+EventTypesListSortFilterProxyModel::EventTypesListSortFilterProxyModel(QObject* parent)
+    : QSortFilterProxyModel(parent) {}
+
+bool EventTypesListSortFilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& /*source_parent*/) const {
+  return sourceModel()->data(sourceModel()->index(source_row, 0), Qt::UserRole).toBool();
+}
+
+bool EventTypesListSortFilterProxyModel::lessThan(const QModelIndex& source_left,
+                                                  const QModelIndex& source_right) const {
+  //Try to sort alpha but keep grouped events together
```
RobertBColton, line 14: That's known as a stable sort.
https://en.wikipedia.org/wiki/Category:Stable_sorts

Models/EventTypesListSortFilterProxyModel.cpp
```diff
@@ -0,0 +1,24 @@
...
+  QString leftGroup = sourceModel()->data(source_left, EventTypesListModel::UserRoles::EventGroupRole).toString();
+  QString rightGroup = sourceModel()->data(source_right, EventTypesListModel::UserRoles::EventGroupRole).toString();
+
+  if (leftGroup == rightGroup) {
+    return sourceModel()->data(source_left, Qt::DisplayRole).toString() <
+           sourceModel()->data(source_right, Qt::DisplayRole).toString();
+  }
+
+  return leftGroup < rightGroup;
```
RobertBColton, line 23: This is an alphanumeric sort, GM and LGM used sorted by the main id, which is why you have alarms and other events appearing before the create event.
Alphanumeric Event Sort Alarms Before Create

Dialogs/EventArgumentsDialog.cpp
fundies: Because they're strings... Arguments are "object", "integer" or "string"

Dialogs/EventArgumentsDialog.cpp
fundies: Clearly I did it because it didn't work. a) your original code was never triggered for spinboxes and b) .toString() changed 33 to "!"

Models/EventTypesListSortFilterProxyModel.cpp
fundies: We do not store integer ids for events anymore so if you want a specific ordering you will need to hard code it in the sort

RobertBColton: It's fine, I just wanted to document it, I don't really care. The only thing I might say to do is we should just always put create first, that's the only part of the order that's stuck in my head, that the create is at the top, like a constructor usually is in a class.

Dialogs/EventArgumentsDialog.cpp
RobertBColton: I'm going to look at it before I merge, it looks like it should work the way I had it, I'll have to test.

Dialogs/EventArgumentsDialog.cpp
RobertBColton: That's what's in event data? Strings is what the shared event stuff gives you to work with? I don't know why Josh would do it that way and not give an enum or some other typed way of doing it.

Models/EventTypesListModel.h
```diff
@@ -0,0 +1,30 @@
...
+class EventTypesListModel : public QAbstractListModel {
+ private:
+  EventData* eventData_;
+
+ public:
+  static enum UserRoles {
+    EventTypeRole = Qt::UserRole,
+    EventGroupRole = Qt::UserRole + 1,
+    EventArgumentsRole = Qt::UserRole + 2,
+    EventBareIDRole = Qt::UserRole + 3
```
RobertBColton, line 21: I don't think you need to do this, I think in C++ enums you only have to specify the starting index.

>An enumerator-definition without an initializer gives the enumerator the value obtained by increasing the value of the previous enumerator by one.
https://stackoverflow.com/a/34811571

```suggestion
    EventGroupRole,
    EventArgumentsRole,
    EventBareIDRole
```

Models/EventTypesListSortFilterProxyModel.cpp
```diff
@@ -0,0 +1,24 @@
+#include "EventTypesListSortFilterProxyModel.h"
+#include "EventTypesListModel.h"
+
+
+EventTypesListSortFilterProxyModel::EventTypesListSortFilterProxyModel(QObject* parent)
+    : QSortFilterProxyModel(parent) {}
+
+bool EventTypesListSortFilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& /*source_parent*/) const {
+  return sourceModel()->data(sourceModel()->index(source_row, 0), Qt::UserRole).toBool();
```
RobertBColton, line 9: You're using a plain `UserRole` here instead of one of the enums you defined.

PR #156, opened 2020-08-22
Background Editor Layout Improvements

RobertBColton:
Bringing some of my layout improvements from #144 over to master as their own changes.

* Changes properties layout from vbox to form so it has two columns where that's needed like the name field.
* Adds the name label and a buddy field.
* Maps the name field to the tree node.
* Changes the tileset groupbox to fixed vertical size policy.

| Master                                                                                                                                             | Pull                                                                                                                                                     |
|----------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
| Master Background Editor Properties Layout | Pull Request Background Editor Properties Layout |

What's odd is the layout looks different on master in the designer than it does at runtime. At runtime the editor above looks like there is a vertical spacer inside the tileset groupbox pushing them all to the top.
Master Background Editor Designer Properties Layout

PR #157, opened 2020-08-23
Sprite Editor Layout Improvements

RobertBColton:
Same as #156 but for the sprite editor.

* Changes properties layout from vbox to form so it has two columns where that's needed like the name field.
* Adds the name label and a buddy field.
* Maps the name field to the tree node.
* Stretches the x/y origin line edits in the parent grid layout.
* Gives the root widget layout a proper "mainLayout" name.
* Gives the origin group layout a proper "originLayout" name.
* Gives the collision shape group layout a proper "collisionShapeLayout" name.
* Gives the BBox group layout a proper "bboxLayout" name.
* Gives all the labels in the BBox dimensions grid the correct buddy.
* Removes the bottom spacer in the properties layout that was used to push everything to the top since form layout does that automatically.

| Master                                                                                                                                             | Pull                                                                                                                                                     |
|----------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
| Master Sprite Editor Properties Layout | Pull Request Sprite Editor Properties Layout |

### DIFF
--- a/Editors/SpriteEditor.ui
+++ b/Editors/SpriteEditor.ui
@@ -169,9 +169,6 @@
             <property name="text">
              <string>&amp;X</string>
             </property>
-            <property name="alignment">
-             <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-            </property>
             <property name="buddy">
              <cstring>originXSpinBox</cstring>
             </property>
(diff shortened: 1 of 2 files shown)


Editors/SpriteEditor.ui
```diff
@@ -127,6 +202,9 @@
             <property name="text">
              <string>&amp;Y</string>
             </property>
+            <property name="alignment">
+             <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
```
JoshDreamland, line 206: I liked these better left-aligned

RobertBColton: I have to agree with you actually, I'll think about it before I merge. As a matter of technicality though, they were right aligned in GM.
GM BBox Labels

RobertBColton: Alright I sent 71843382a789008d6c98d9feb5250389508bf849 to put it back for now.

PR #163, opened 2020-08-26
Included File Editor

RobertBColton:
This adds an editor for included files like the one I added to LateralGM which I based off of GameMaker 5's data file editor.

* We still need an include proto, so I haven't added any actual functionality and you can't open the editor yet.
* I decided to continue with the file name being separate, like GM5 and _unlike_ GMS, so we don't have file extensions in the tree.
* I've clearly kept several options that GMSv1.4 deprecated which I am not sure if are still useful.
* I've not included the target options that GMSv1.4 introduced because I don't know how we plan to handle that yet.

| RGM | LGM | GM5 | GMSv1.4 |
|-----|-----|-----|---------|
|RGM Include Editor|LGM Include Editor|GM5 Include Editor|GMS Include Editor|

### DIFF
--- a/Editors/IncludeEditor.cpp
+++ b/Editors/IncludeEditor.cpp
@@ -1,0 +1,14 @@
+#include "IncludeEditor.h"
+#include "ui_IncludeEditor.h"
+
+IncludeEditor::IncludeEditor(QWidget *parent) :
+  QWidget(parent),
+  ui(new Ui::IncludeEditor)
+{
+  ui->setupUi(this);
+}
+
+IncludeEditor::~IncludeEditor()
+{
+  delete ui;
+}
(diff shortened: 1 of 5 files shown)


Editors/IncludeEditor.cpp
```diff
@@ -0,0 +1,14 @@
+#include "IncludeEditor.h"
+#include "ui_IncludeEditor.h"
+
+IncludeEditor::IncludeEditor(QWidget *parent) :
+  QWidget(parent),
+  ui(new Ui::IncludeEditor)
```
JoshDreamland, line 6: Why isn't this some kind of smart pointer?

RobertBColton: Qt autogenerated that from the class creator and Qt objects use the Qt ownership semantics.
https://doc.qt.io/qt-5/objecttrees.html

If you're proposing we make it a smart pointer, we can, but I'm not so sure it's a good idea we go against the Qt conventions. We might shoot ourselves in the foot that way.

PR #165, opened 2020-08-27
Information Editor

RobertBColton:
This adds an editor for the old game information resource or some derivative of it. I am not proposing we merge this instead of deprecating it like GMS2 anytime soon, but it's on the table once we do make a decision. The editor is really simple since it's just essentially Microsoft WordPad and Qt has an official example for it. I've already done most of the logic here and the UI is completely and correctly synchronized to state changes in the document and its formatting.

I've maybe somewhat changed my opinion about this resource, though I've not swayed Josh on it. Rather than deprecate it like YoYoGames, I think it could maybe still be useful for game rendered dialogs and such. Styled text is a feature a lot of game engines surprisingly don't actually support. It's not only friendly to new users, it compresses better than a large prerender, and it's easier to change later on. I would probably not think it a good idea we continue using RTF, but instead use something akin to Markdown for EGM and the engine. I would like to see something like `draw_info` rather than `show_info` that makes the text an actual integrated part of the game.

* We need a bulleted list icon in Calico for this editor. We could provide numbered lists too.
* We need a special icon for font and text color since I reused the font resource icon for both until then.
* The edit menu is still empty until I decide to reuse the text edit's default context menu or define custom cut/copy/paste actions so the keybinding preferences can pick them up.
* I did not add a status bar just like LGM didn't since GM never actually used the one it had and it just wastes space?
* I mapped load file to `CTRL+O` instead of `File->Options` like LGM because GM is inconsistent with that. Most programs and even GM in its main top-level File menu use `CTRL+O` as the open a file shortcut.
* Mnemonics and shortcuts are otherwise the same as GM and LGM where I could add them.

| RGM | LGM | GM8.1 | GMSv1.4 |
|-----|-----|-----|---------|
|RGM Information Editor|LGM Information Editor|GM8.1 Information Editor|GMS Information Editor|


RobertBColton:
Merging this as it can later be repurposed into a markdown editor. Also there's no reason the engine can't support rich text rendering through an extension.

PR #166, opened 2020-08-28
Path Editor Points Display Model

RobertBColton:
This adds a proxy model around the path points repeated model for display purposes only. It allows us to add additional icons and decorative crap to the path points table. In other words, if you break it, you buy it.

* I did not solve #127 here because I don't want to go overboard yet so you still are not able to see the speed column.
* I inlined the proxy model class into the path editor source because I don't want or expect it to get very big.
* I tried doing this with CSS first but it can't be done because Qt lacks an `nth-child` for the `::section` selector.
https://doc.qt.io/qt-5/stylesheet-reference.html
* The three RGB diamond icons are a 0.5 zoom and saturation shift of some diamond icon Josh just added to Calico.
* The motion icon for the speed column is the media speed icon from Calico with the background stripped off.
* This could be done in the base repeated model if we add more reflective stuff to reflect upon and choose an icon.
* This can not be done with an item delegate because the header views don't support those.
>Note: Each header renders the data for each section itself, and does not rely on a delegate. As a result, calling a header's setItemDelegate() function will have no effect.
https://doc.qt.io/qt-5/qheaderview.html#appearance
* This could be done instead by making a custom QHeaderView and painting the header ourselves. That is also the only way to get icon alignment if we wanted the icon centered or aligned with the text.

Path Editor Points Table Header


RobertBColton:
Closing in favor of #167 which I just merged.

PR #167, opened 2020-08-28
Proto Model Header Data Mutation

RobertBColton:
This is an alternative to #166 that adds the ability to mutate the header data returned by the base proto model. I use it to give the column headers of the path editor's points list special icons. I hid the vertical header of the path points list table too because we don't actually want it and it started showing row numbers now that my implementation allows the base `QAbstractItemModel::headerData` to return the section number as a string for `Qt::DisplayRole` role. We can also use this later to translate the column headers displaying field names for i18n.

I was also thinking that this could be static and fundies did mention I could just use the proto type name so that messages still get unique header data. However, I don't know what other kind of dynamic crap we might want to do so I decided not to make it static yet. Static doesn't exactly work with things like sorting and other transformations of the data.


RobertBColton:
I didn't do that fundies because I want to be able to set the tooltip and stuff too. In the room layers view I'm just going to show eyeball icon for layer visiblity with a "Visible" tooltip.

PR #172, opened 2020-08-29
Sprite Show CheckBox Labels

RobertBColton:
This fixes the show checkboxes on the sprite form so you can click the label to toggle the checkbox. I removed the second label and added the text directly to the checkbox. Then I used `setLayoutDirection` to put the label to the left of the checkbox like fundies had it. This does not affect the text layout direction according to the documentation so it should not interfere with i18n here.
https://doc.qt.io/qt-5/qwidget.html#layoutDirection-prop


RobertBColton:
No, I have no idea why fundies wanted to put the label before the checkbox, but that's why it's so messed up. My other pull request moves this shit to the XML. Regardless, the way fundies did it with a separate label makes the label not toggle the checkbox. Arabic wouldn't work the way fundies had it either.

PR #177, opened 2020-08-31
Simpler Path Point Row Removal

RobertBColton:
Why use the row removal operation inline when you don't have to? I can see why the model would use it to implement remove rows but not why it's used outside the model. I can do the same thing the 4 locs are doing in a single loc and it works fine!


RobertBColton:
Closing because I see that this is Josh's solution to that quadratic behavior I was bitching about. We could have actually still merged this though because it's not really necessary when removing a single row with [QAbstractItemView::SingleSelection](https://doc.qt.io/qt-5/qabstractitemview.html#SelectionMode-enum) but I would rather promote the use of the class to avoid somebody accidentally hitting the quadratic runtime. I am going to follow this up with a pull request that accepts QModelIndexList for deleting selected indexes in bulk avoiding the duplicating behavior I am currently bitching about.

PR #189, opened 2020-09-03
Path Point Fields Translation

RobertBColton:
This uses the header data override of #175 to translate the path point list column headers.

|Master|Pull|
|----|----|
|Path Point Columns Master|Path Point Columns Pull|


JoshDreamland (approved):
Probably worth writing a wrapper that binds the field and translates its qualified name.

RobertBColton:
That was actually what I said to fundies when we first added this, that the mutable header data could be static mapped to the message type name. I know by qualified name you are referring to the field's full name with the message type name prefixed to the field name. fundies said we could actually do both too, and have the non static header data override the static header data. I decided to just do the non static behavior to start with because I wasn't sure, but what you are proposing could work I'm sure.
FieldDescriptor Fully Qualified Name

PR #190, opened 2020-09-05
Font Preview Initialize

RobertBColton:
We can't set the preview font on the preview text area _before_ we've mapped the property widgets to the model. We have to grab the values out of the model to build a QFont, which is easier to grab from the mapped and loaded widgets. We also need to set the initial font on the range preview too. The previews now have the correct font the first time the editor is opened.


JoshDreamland (approved):
I assume _font is not a proto, and that's why our font infrastructure doesn't apply to it. I think I'd be more comfortable if it worked with the proto instead, but okay.

PR #191, opened 2020-09-06
Static Header Data

RobertBColton:
This adds the static header data I mentioned in #167 and in #189 Josh suggested. It turned out to be more complicated than I thought due to initialization constraints, but it essentially worked out the way me and Josh expected. I am not sure if we actually want to go with it, but it does work. The path editor is still being translated and showing icons in header data.

Static Path Editor Header Data

### DIFF
--- a/Editors/PathEditor.cpp
+++ b/Editors/PathEditor.cpp
@@ -10,6 +10,19 @@
 #include <QSpinBox>
 #include <QToolButton>
 
+void initialize_static_path_header_data() {
+  auto descriptor = Path::Point::GetDescriptor();
+  auto x_name = descriptor->FindFieldByNumber(Path::Point::kXFieldNumber)->full_name();
+  auto y_name = descriptor->FindFieldByNumber(Path::Point::kYFieldNumber)->full_name();
+  auto speed_name = descriptor->FindFieldByNumber(Path::Point::kSpeedFieldNumber)->full_name();
+  ProtoModel::SetHeaderData(x_name, Qt::Horizontal, PathEditor::tr("X"), Qt::DisplayRole);
+  ProtoModel::SetHeaderData(y_name, Qt::Horizontal, PathEditor::tr("Y"), Qt::DisplayRole);
+  ProtoModel::SetHeaderData(speed_name, Qt::Horizontal, PathEditor::tr("Speed"), Qt::DisplayRole);
+  ProtoModel::SetHeaderData(x_name, Qt::Horizontal, QIcon(":/actions/diamond-red.png"), Qt::DecorationRole);
+  ProtoModel::SetHeaderData(y_name, Qt::Horizontal, QIcon(":/actions/diamond-green.png"), Qt::DecorationRole);
+  ProtoModel::SetHeaderData(speed_name, Qt::Horizontal, QIcon(":/actions/motion.png"), Qt::DecorationRole);
+}
+
 PathEditor::PathEditor(MessageModel* model, QWidget* parent) : BaseEditor(model, parent), _ui(new Ui::PathEditor) {
   _ui->setupUi(this);
   connect(_ui->saveButton, &QAbstractButton::pressed, this, &BaseEditor::OnSave);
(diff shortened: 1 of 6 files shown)


main.cpp
```diff
@@ -21,6 +22,8 @@ int main(int argc, char *argv[]) {
   a.setWindowIcon(QIcon(":/icon.ico"));
   a.setAttribute(Qt::AA_DisableWindowContextHelpButton);
 
+  initialize_static_path_header_data();
```
RobertBColton, line 25: If this isn't called after a QApplication exists, it will crash trying to construct a QIcon or QPixmap because the icon engine isn't initialized. That's why I couldn't just do a static object initialization.

main.cpp
```diff
@@ -1,6 +1,7 @@
 #include "main.h"
 #include "MainWindow.h"
 
+#include "Editors/PathEditor.h"
```
RobertBColton, line 4: Where would we go with all this initialization? I wanted it to be in the editor classes. We would need to make a more refined API for custom editors too. What if we want to do dynamic translations at runtime later.

Models/RepeatedMessageModel.h
```diff
@@ -31,6 +31,9 @@ class RepeatedMessageModel : public RepeatedModel<Message> {
   virtual QVariant data(const QModelIndex &index, int role) const override;
   virtual int columnCount(const QModelIndex &parent = QModelIndex()) const override;
   virtual Qt::ItemFlags flags(const QModelIndex &index) const override;
+  virtual const Descriptor* GetDescriptor() const override {
+    return _field->message_type();
+  }
```
RobertBColton, line 36: This is fragile as all hell.

Models/ProtoModel.h
```diff
@@ -110,6 +115,8 @@ class ProtoModel : public QAbstractItemModel {
   bool _dirty;
   Message *_protobuf;
   ProtoModel *_parentModel;
+  static QHash<QString,QHash<Qt::ItemDataRole,QVariant>> _horizontalProtoHeaderData;
+  static QHash<QString,QHash<Qt::ItemDataRole,QVariant>> _verticalProtoHeaderData;
```
RobertBColton, line 119: I was going to use `std::string` as the key type, but QHash doesn't have a hash function for that. If we used `std::string` we wouldn't have to unnecessarily box and unbox the key as a QString, since this is our API and we're the only ones calling it. I would have used `std::unordered_map` but I didn't want to be inconsistent with the existing header data map.

Editors/PathEditor.cpp
```diff
@@ -10,6 +10,19 @@
...
+  auto x_name = descriptor->FindFieldByNumber(Path::Point::kXFieldNumber)->full_name();
+  auto y_name = descriptor->FindFieldByNumber(Path::Point::kYFieldNumber)->full_name();
+  auto speed_name = descriptor->FindFieldByNumber(Path::Point::kSpeedFieldNumber)->full_name();
+  ProtoModel::SetHeaderData(x_name, Qt::Horizontal, PathEditor::tr("X"), Qt::DisplayRole);
+  ProtoModel::SetHeaderData(y_name, Qt::Horizontal, PathEditor::tr("Y"), Qt::DisplayRole);
+  ProtoModel::SetHeaderData(speed_name, Qt::Horizontal, PathEditor::tr("Speed"), Qt::DisplayRole);
+  ProtoModel::SetHeaderData(x_name, Qt::Horizontal, QIcon(":/actions/diamond-red.png"), Qt::DecorationRole);
+  ProtoModel::SetHeaderData(y_name, Qt::Horizontal, QIcon(":/actions/diamond-green.png"), Qt::DecorationRole);
+  ProtoModel::SetHeaderData(speed_name, Qt::Horizontal, QIcon(":/actions/motion.png"), Qt::DecorationRole);
+}
```
RobertBColton, line 24: Here it is in all its glory, the static header data. Note I am using the static QObject translate, which I don't know if it effects dynamic translation of the UI at runtime (say for switching the language dynamically).

RobertBColton:
This is no longer needed as of #201 as Josh has already rewritten the header data into a modular configuration.
https://github.com/enigma-dev/RadialGM/blob/109ea381aca0b7396723c38e701b424b1f9ddda2/MainWindow.cpp#L404
However, what is still needed is better organization of this to allow for plugins and custom editors as mentioned above.

PR #192, opened 2020-09-06
Git UI

RobertBColton:
WIP, not done yet. Can't find a way to tabify the docks in the designer, had to do it in code. I think later we might split some of these docks out into their own UI files.

image


RobertBColton:
Closed in favor of #198 from the same branch.

PR #197, opened 2020-11-17
fix build

fundies:


Editors/BackgroundEditor.cpp
```diff
@@ -76,11 +76,11 @@ void BackgroundEditor::on_actionLoadImage_triggered() {
 
   if (dialog->exec() && dialog->selectedFiles().size() > 0) {
     QString fName = dialog->selectedFiles()[0];
-    if (fName.endsWith("Background.gmx")) {
-      Background* bkg = gmx::LoadBackground(fName.toStdString());
-      if (bkg != nullptr) {
+    if (fName.endsWith("Background.gmx") || fName.endsWith(".bkg")) {
```
JoshDreamland, line 79: Are these checks case-insensitive? The capital B suggests not.

JoshDreamland (approved):

Editors/BackgroundEditor.cpp
fundies: these checks need work but jus tryna fix build here

PR #198, opened 2020-11-21
Git ui /egm integration WIP

fundies:


Components/EGMManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+ResourceChangesModel& EGMManager::GetResourceChangesModel() {
+  return _resChangeModel;
+}
+
+GitHistoryModel& EGMManager::GetGitHistoryModel() {
+  return _gitHistoryModel;
+}
+
+void EGMManager::ResourceChanged(Resource& res, ResChange change, const QString& oldName) {
```
fundies, line 60: this method I need help with and is very incomplete

Models/ResourceChangesModel.h
```diff
@@ -0,0 +1,32 @@
...
+
+enum class ResChange : int {
+  Added,
+  Modified,
+  Renamed,
+  Removed,
+  Reverted
+};
+
+class ResourceChangesModel : public QAbstractListModel
```
JoshDreamland, line 18: Is it a common practice to create separate models to use only as an index over some display model? My understanding is that we generally don't create a model for something we don't intend to display in the UI. And you're not planning to display all resources as a list, right? If you're just adding this model class as something that listens for resource changes and updates disk, I don't think that's right. I think it can just be a normal manager class. @RobertBColton please confirm or deny.

Models/GitHistoryModel.cpp
```diff
@@ -0,0 +1,83 @@
...
+int GitHistoryModel::columnCount(const QModelIndex &parent) const {
+  if (parent.isValid()) return 0;
+  return 5;
+}
+
+QVariant GitHistoryModel::headerData(int section, Qt::Orientation orientation, int role) const {
+  if (role != Qt::DisplayRole || orientation != Qt::Orientation::Horizontal) return QVariant();
+
+  switch(section) {
+    case 0: return tr("Tree");
```
JoshDreamland, line 60: I don't know a better way of doing this, but I feel like there probably is one... I mean, naively, you can just put those labels in an array. But I'm not sure what happens if the user tries to drag columns around or something; I feel like Qt probably has better support for this somewhere.

MainWindow.h
```diff
@@ -27,19 +26,35 @@ class MainWindow;
 class MainWindow : public QMainWindow {
   Q_OBJECT
 
+ private:
```
JoshDreamland, line 29: This move confused me for a second. As a style rule, we (primarily Google) put the private block at the bottom of the class (so that things we want people to know about are on top).

Components/GitTreeStyledDelegate.h
```diff
@@ -0,0 +1,15 @@
+#ifndef GITTREESTYLEDDELEGATE_H
```
JoshDreamland, line 1: File comment: I'd move the delegates and various Git list facets to a subfolder of this components folder, to improve navigability of the codebase. It's honestly a big feature, which is why I thought it'd make a good intern project.

Models/ResourceChangesModel.h
fundies: this is displayed tho

PR #199, opened 2020-11-22
Add TreeModel::MakeResourcePath and ResourceMoved.

JoshDreamland:
These should enable moving resources on disk in response to moving that resource in the tree.

We may need to revisit the TreeModel to make it use a custom wrapper of the TreeNode proto. This would save us some headache keeping track of the tree structure, and also where the files live on disk, when they were last saved, etc.

### DIFF
--- a/Models/TreeModel.cpp
+++ b/Models/TreeModel.cpp
@@ -158,6 +158,18 @@ int TreeModel::rowCount(const QModelIndex &parent) const {
   return parentItem->child_size();
 }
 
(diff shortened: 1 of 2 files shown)


RobertBColton (approved):
It "looks" correct; it's what you're trying to do. I'm still not sure it's actually necessary though, but I'd probably suggest a better alternative once I see how it's actually used. Also, programmatic changes in the model are not reflected by this change.

JoshDreamland:
We need to grow this model out to be *the* authoritative source for that.

PR #201, opened 2021-01-02
New TreeModel [WIP]

JoshDreamland:
This commit breaks almost everything.

Redoes the ProtoModel family of classes to not know about our specific protos. Disclaimer: some references may still exist, but are unnecessary for the resource tree to function. If I missed any references, they're used by, eg, one of the editor classes.

I am basically finished modifying the Model classes and need to start updating the editors to use zero-based field indexing rather than field numbers. I believe Robert has done this work already, so I decided to stop where I was and create this pull request.

I believe we should be able to consolidate this with Robert's PR and have something pretty close to the finished product.

Currently broken:
- Most editors are broken in some way.
  - Paths crash
  - Sprites can't update BBox parameters
  - Rooms don't display anything
  - Objects display very little.
- A lot of signals are now disconnected (resource renames, for example).
- This system might allow models to be null under some circumstances.
  - Some messages can recursively contain messages to infinity, so we can't build out the entire model. Somewhere in there, something comes out null, and I can't remember what right now.


Models/TreeModel.cpp
```diff
@@ -689,13 +689,13 @@ QModelIndex TreeModel::addNode(const Message &child, const QModelIndex &parent)
...
-    return node->mapFromSource(node->repeated_model->duplicate(mapToSource(index)));
-  }*/
-  qDebug() << rowCount() << " rows";
-  return QModelIndex();
+QModelIndex TreeModel::duplicateNode(const QModelIndex & index) {
+  Node *node = IndexToNode(index);
+  if (node->IsRepeated()) {
+    //return node->mapFromSource(node->repeated_model->duplicate(mapToSource(index)));
+  }
+  Node *parent = IndexToNode(index.parent());
```
JoshDreamland, line 698: PLEASE check these; each one of these is capable of returning null, and the safety checks I added only work if you actually verify that I didn't return null.

Models/TreeModel.cpp
```diff
@@ -689,13 +689,13 @@ QModelIndex TreeModel::addNode(const Message &child, const QModelIndex &parent)
...
-QModelIndex TreeModel::duplicateNode(const QModelIndex & /*index*/) {
-  /*Node *node = IndexToNode(index);
-  if (node->repeated_model) {
-    return node->mapFromSource(node->repeated_model->duplicate(mapToSource(index)));
-  }*/
-  qDebug() << rowCount() << " rows";
-  return QModelIndex();
+QModelIndex TreeModel::duplicateNode(const QModelIndex & index) {
+  Node *node = IndexToNode(index);
+  if (node->IsRepeated()) {
```
JoshDreamland, line 695: Like this change. But please just leave the if() commented if it's not being used.

fundies (approved):

PR #206, opened 2021-03-28
Update main.cpp

ThanoS-max:


JoshDreamland:
Hi Aditya,

Thanks for your interest, but we follow a version of [Google C++ style](https://google.github.io/styleguide/cppguide.html), so this change is actually inconsistent with our preferred style. Spaces are only included around `->` when indicating a return type (not when accessing a pointer).

If you're looking for ways to get started, it might be easier to jump in on [the engine repo](/enigma-dev/enigma-dev).

Cheers

PR #208, opened 2021-09-13
Update room editor on proto changes

fundies:


JoshDreamland (approved):
Looks objectively correct, thanks

PR #210, opened 2021-09-19
fix submodule url && emake server shutdown

fundies:


Plugins/ServerPlugin.cpp
```diff
@@ -330,11 +338,14 @@ ServerPlugin::ServerPlugin(MainWindow& mainWindow) : RGMPlugin(mainWindow) {
   compilerClient->GetSystems();
 }
 
-ServerPlugin::~ServerPlugin() { process->close(); }
+ServerPlugin::~ServerPlugin() {
+  compilerClient->TearDown();
+  process->waitForFinished(1000);
```
JoshDreamland, line 343: Is this one second? Honestly, the default of 30 (or maybe 10 or 15) seems fine. Sometimes there will be outstanding requests. I'm really not comfortable with anything under five.

JoshDreamland (approved):

Plugins/ServerPlugin.cpp
fundies: k

PR #215, opened 2021-10-23
Path Editor/Field Label Localization

RobertBColton:
This adds the ability to localize the field name from #191 in addition to the field icon in the column headers to Josh's version.

Path Editor Point Field Localization

GameMaker Studio 2 also has this for the path editor:
https://manual.yoyogames.com/The_Asset_Editors/Paths.htm
GMS2 Path Editor Points

### DIFF
--- a/Models/ProtoModel.cpp
+++ b/Models/ProtoModel.cpp
@@ -70,8 +70,8 @@ void ProtoModel::DisplayConfig::SetMessageLabelField(const std::string &message,
 }
 void ProtoModel::DisplayConfig::SetMessageIconIdField(const std::string &message, const FieldPath &field_path,
                                                       FieldDisplayConfig::IconLookupFn icon_lookup_function) {
-  message_display_configs_[message].icon_field = field_path;
   std::string field = field_path.str();
+  message_display_configs_[message].icon_field = field_path;
   if (field_path.size() != 1)
     qDebug() << "Warning: Nested icon fields not currently implemented; `" << field.c_str() << "` won't work properly";
   field_display_configs_[field].icon_lookup_function = icon_lookup_function;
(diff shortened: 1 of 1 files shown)


Models/ProtoModel.cpp
```diff
@@ -89,14 +88,20 @@ void ProtoModel::DisplayConfig::SetFieldDefaultIcon(const FieldDescriptor *field
 
 void ProtoModel::DisplayConfig::SetFieldHeaderIcon(const std::string &message, const FieldPath &field_path,
                                                    const QString &icon_name) {
-  std::string field = message;
-  message_display_configs_[message].icon_field = field_path;
```
RobertBColton, line 93: Pretty sure this wasn't supposed to be in this method? The `SetMessageIconIdField` already covers its behavior.

Models/ProtoModel.cpp
```diff
@@ -97,6 +97,16 @@ void ProtoModel::DisplayConfig::SetFieldHeaderIcon(const std::string &message, c
   field_display_configs_[field].header_icon = icon_name;
 }
 
+void ProtoModel::DisplayConfig::SetFieldHeaderLabel(const std::string &message, const FieldPath &field_path,
+                                                   const QString &label) {
+  std::string field = message;
+  message_display_configs_[message].icon_field = field_path;
+  for (const auto &fcomp : field_path.fields) field += "." + fcomp->name();
+  if (field_path.size() != 1)
+    qDebug() << "Warning: Nested label fields not currently implemented; `" << field.c_str() << "` won't work properly";
```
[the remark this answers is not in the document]
RobertBColton: Ok, let's just do the full name then: 73a0323cf0e1cefbd4dfebae5a8f8a7971dde6ff

Models/ProtoModel.cpp
RobertBColton: I see the other one is actually specifying a lambda lookup, so maybe that's why.
I think there's more permutations of these functions that can be added later actually.
I put them back for now and decided to punt: 0c935917fcf42f6990fa3730ba69e08840d125a1

Issue #223, opened 2022-03-31
Automatic resource naming seems inconsistent


KartikShrivastava:
Hi,
I've noticed that whenever we create a new resource under same resource type folder, the index in the name does not increment automatically. It would be convinient to increment the index if user relies on automatic naming.
For example: These three sprite resources are created with the same name:
![image](https://user-images.githubusercontent.com/26518513/161004185-0a25578b-8e51-4ca3-b45e-dc4b8a207446.png)

RobertBColton:
It's supposed to, something must have broken.

PR #226, opened 2022-04-17
Fix for naming index bug #223

KartikShrivastava:
A small fix for #223 which involves addition of `resourceMap->AddResource(...)` call after resource name creation `resourceMap->CreateResourceName`. This call helps to increment index counter when another resource of same type is created.

### DIFF
--- a/MainWindow.cpp
+++ b/MainWindow.cpp
@@ -564,7 +564,13 @@ void MainWindow::CreateResource(TypeCase typeCase) {
   refl->MutableMessage(&child, field);
 
   // find a unique name for the new resource
-  child.set_name(resourceMap->CreateResourceName(&child).toStdString());
+  QString resourceName = resourceMap->CreateResourceName(&child);
+  child.set_name(resourceName.toStdString());
+
+  // add new resource with created name, helps in creating another unique name
+  auto messageModelRef = MessageModel(ProtoModel::NonProtoParent{this}, &child);
+  resourceMap->AddResource(typeCase, resourceName, &messageModelRef);
+
   // release ownership of the new child to its parent and the tree
   auto index = this->treeModel->addNode(child, _ui->treeView->currentIndex());
   treeModel->triggerNodeEdit(index, _ui->treeView);


fundies:
> A small fix for #223 which involves addition of `resourceMap->AddResource(...)` call after resource name creation `resourceMap->CreateResourceName`. This call helps to increment index counter when another resource of same type is created.

Did you check if other cases like rename and delete are properly updating that map?

KartikShrivastava:
> Did you check if other cases like rename and delete are properly updating that map?

Hi, I just checked for delete and indeed `_resources` in `ResourceModelMap` is not updating properly.
Here is what happens right now: After creating sprite0, sprite1 and sprite2, if I delete sprite1 and create another sprite, it is named sprite3. Which ig means that sprite1 is still there in `_resources[sprite]`.

Rename is working properly.

I'll fix delete and make sure to test properly before creating new PRs.

KartikShrivastava:
Hi, I've tested correctness of resource name_index by both deleting and updating resources for 3-4 resource types. It's working fine now.

Models/ResourceModelMap.cpp
```diff
@@ -129,8 +129,9 @@ void ResourceModelMap::ResourceRemoved(TypeCase type, const QString& name,
     }
   }
 
-  // Remove an references to this resource
-  emit ResourceRenamed(ResTypeAsString(type), name, "");
```
JoshDreamland, line 133: Do we have slots listening on this signal? If so, killing `ResourceRenamed` might break something (e.g. the EGM handler might want to know that this happened so it can issue a rename on the underlying files). It might be worth emitting both signals.

That said, this is probably underthought—that rename might fail (e.g. the resulting path might be too long or the name might conflict with another entity).

MainWindow.cpp
```diff
@@ -564,7 +564,13 @@ void MainWindow::CreateResource(TypeCase typeCase) {
   refl->MutableMessage(&child, field);
 
   // find a unique name for the new resource
-  child.set_name(resourceMap->CreateResourceName(&child).toStdString());
+  QString resourceName = resourceMap->CreateResourceName(&child);
+  child.set_name(resourceName.toStdString());
+
+  // add new resource with created name, helps in creating another unique name
+  auto messageModelRef = MessageModel(ProtoModel::NonProtoParent{this}, &child);
```
JoshDreamland, line 571: Why `auto messageModelRef = MessageModel(...)` over `MessageModel messageModelRef{...}`?

MainWindow.cpp
```diff
@@ -564,7 +564,13 @@ void MainWindow::CreateResource(TypeCase typeCase) {
   refl->MutableMessage(&child, field);
 
   // find a unique name for the new resource
-  child.set_name(resourceMap->CreateResourceName(&child).toStdString());
+  QString resourceName = resourceMap->CreateResourceName(&child);
+  child.set_name(resourceName.toStdString());
+
+  // add new resource with created name, helps in creating another unique name
+  auto messageModelRef = MessageModel(ProtoModel::NonProtoParent{this}, &child);
+  resourceMap->AddResource(typeCase, resourceName, &messageModelRef);
```
JoshDreamland, line 572: `messageModelRef` is a temporary in this context; this reference will die after this call to `CreateResource` terminates. Doesn't this cause instability in the IDE?

Models/ResourceModelMap.cpp
```diff
@@ -129,8 +129,9 @@ void ResourceModelMap::ResourceRemoved(TypeCase type, const QString& name,
...
+  // Remove references to this resource
+  _resources[type].remove(name);
+  emit DataChanged();
```
JoshDreamland, line 134: Okay. I understand what this is for, now; the original line was a hack. The main purpose of the ResourceRenamed call, from that comment, seems to have been to update code and resource references to the resource. So if you delete `sprite0` but `object0` had it set as its own sprite, the reference is updated to be the empty string. I guess we can assume that `DataChanged` will accomplish that, but be prepared to spot this regression.

MainWindow.cpp
```diff
@@ -564,10 +564,19 @@ void MainWindow::CreateResource(TypeCase typeCase) {
...
-  child.set_name(resourceMap->CreateResourceName(&child).toStdString());
+  QString resourceName = resourceMap->CreateResourceName(&child);
+  child.set_name(resourceName.toStdString());
+
   // release ownership of the new child to its parent and the tree
   auto index = this->treeModel->addNode(child, _ui->treeView->currentIndex());
   treeModel->triggerNodeEdit(index, _ui->treeView);
+
+  // add new resource with created name, helps in creating another unique name
+  MessageModel* messageModel = treeModel->IndexToNode(index)->BackingModel()->TryCastAsMessageModel();
```
JoshDreamland, line 575: This chain actually has room for a bunch of nullptrs in it. I believe both IndexToNode and BackingModel can return null, and having `this` as nullptr is UB (it's legal to optimize out checks for `this == null`).

JoshDreamland (approved):
Thanks! I think we're good to go. Anything else from you, fundies?

PR #229, opened 2022-06-07
Add egm file save functionality

KartikShrivastava:


MainWindow.cpp
```diff
@@ -469,8 +469,35 @@ void MainWindow::on_actionOpen_triggered() {
...
+
+    auto itr = extensionMap.begin();
+    QString f = itr->first;
+    ++itr;
+
+    for(; itr != extensionMap.end(); ++itr){
+      f += ";;" + itr->first;
+    }
+    return f;
+  };
```
bjorn, line 489: If you use `QHash` instead of `std::unordered_map` you can just write:
```c++
const auto filters = extensionMap.keys().join(QStringLiteral(";;"));
```
I'm sure something is possible with STL, but in general Qt containers provide a little more convenience. :-)

However, why do we do this given that we support only one file format anyway? Are additional file formats planned for the Save function?

MainWindow.cpp
```diff
@@ -469,8 +469,35 @@ void MainWindow::on_actionOpen_triggered() {
...
+
+  QString selectedFilter;
+  QString fileName = QFileDialog::getSaveFileName(this, tr("Save Project"), "", getFilters(), &selectedFilter);
+
+  // to check if fileName contains extension
+  const std::filesystem::path &filePath = fileName.toStdString();
+  const std::string &ext = ToLower(filePath.extension().u8string());
+  if (ext.empty()) {
```
bjorn, line 497: All this conversion just to check the extension? And why lowercasing for empty check? If we stay with Qt API, we just do:
```c++
if (QFileInfo(fileName).suffix().isEmpty()) {
```
However, this will still fail to add an extension to any files with a dot in their name, or ending with a dot. I think we just want to make sure to add the extension if needed. Maybe something like:
```c++
if (!fileName.endsWith(extensionMap[selectedFilter], Qt::CaseInsensitve)) {
    while (fileName.endsWith(QLatin1Char('.'))
        fileName.chop(1);
    fileName.append(extensionMap[selectedFilter]);
}
```

MainWindow.cpp
KartikShrivastava: I guess I liked qt way of doing that more, didn't know that, chages on way. I'm not sure about supporting extensions in future, just thought a little ahead of time, that maybe someone adds another extension.

MainWindow.cpp
KartikShrivastava: You're right ToLower is not useful here, I copy-pasted it from other part and overlooked the usage. I've made the changes!

MainWindow.cpp
```diff
@@ -468,6 +468,28 @@ void MainWindow::on_actionOpen_triggered() {
...
+                                                  &selectedFilter);
+
+  if(!fileName.endsWith(extensionMap[selectedFilter], Qt::CaseInsensitive)) {
+    // removes any trailing periods(.) from the file path
+    while(fileName.endsWith(QLatin1Char('.')))
+      fileName.chop(1);
+    fileName.append(extensionMap[selectedFilter]);
+  }
+  std::cout<<fileName.toStdString()<<"\n";
+  if(!fileName.isEmpty()) {
```
bjorn, line 488: This check needs to be moved up. Now it will never be false, since we'd have added the extension.

MainWindow.cpp
fundies: Additional formats are possible but not planned as of yet. Someone way want to do gmx export later.

fundies (approved):

fundies (approved):

PR #236, opened 2024-03-17
Fix CMake error because of gRPC and Protobuf

### LINKED ISSUE #235: CMake error due to gRPC
The following error appears when running cmake command:
```bash
~/Desktop/RadialGM/build$ cmake ..
```
```
CMake Error at CMakeLists.txt:248 (find_package):
Found package configuration file:

  /usr/local/lib/cmake/grpc/gRPCConfig.cmake

but it set gRPC_FOUND to FALSE so package "gRPC" is considered to be NOT
FOUND.  Reason given by package:

The following imported targets are referenced, but are missing:
protobuf::libprotobuf protobuf::libprotoc
```

k0T0z:
Fixes #235


k0T0z:
Closing as merged in #237

k0T0z:
Reopened because this error exists on Ubuntu only I think

